PR #111, opened 2013-06-07
AMQP receiver reads metric from body when AMQP_METRIC_NAME_IN_BODY is True

antoinerg:
It seems to me that if AMQP_METRIC_NAME_IN_BODY is True, we should retrieve the name of the metric from the body (and not the other way around).


obfuscurity:
@tld is correct, no issue here.

PR #116, opened 2013-06-23
use category specific logging instead of log.msg()

anomalizer:
lib/carbon/log has helper methods that let you record the identity of
the component that is issuing a log statement. This requires the
application code to invoke log.foo() as opposed to log.msg() where __foo__
can be one of the predefined names such as _relay_, _cache_, _query_ etc. etc.

While large parts of the application do use these more specialized
methods, there exists some part of the code base that does not use these.
This patch fixes up all such invocations


obfuscurity:
I took a look at some of these and I'm not sure they justify new logfiles on their own. FWIW after looking through all of our logging code I :100: agree there needs to be more consistency, but I'd prefer a holistic approach at updating everything.

PR #126, opened 2013-08-04
Simple security for TCP and UDP line receivers.

lochiiconnectivity:
Addition of support for a simple plaintext 'key' in the line and UDP receiver services, the presence of this key acts as a simple means to prevent unauthorised or unwanted metric sources from using these services. For better security, please consider using AMQP.

I'm currently using this in a production environment. I opted not to mangle the metric path.


obfuscurity:
Closing in favor of #441. Please move all discussion over there.

PR #135, opened 2013-08-29
Restore CH routing that replicas exist on different servers

dctrwatson:
Should put a note that 0.9.11+ is not backwards compatible due to this change

### DIFF
--- a/lib/carbon/routers.py
+++ b/lib/carbon/routers.py
@@ -63,12 +63,18 @@ class ConsistentHashingRouter(DatapointRouter):
   def getDestinations(self, metric):
     key = self.getKey(metric)
 
-    for count,node in enumerate(self.ring.get_nodes(key)):
-      if count == self.replication_factor:
+    used_servers = set()
+
+    for (server, instance) in self.ring.get_nodes(key):
+      if server in used_servers:
+        continue
+      else:
+        used_servers.add(server)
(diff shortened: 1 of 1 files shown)


deniszh:
@fhossain: nope, your fix is giving serious performace regression - please check  https://github.com/graphite-project/carbon/pull/175
So, we need to stick with @dctrwatson  fix.

deniszh:
@dctrwatson, could you please explain rationale behind that change, please?

dctrwatson:
It's been awhile.

iirc, the main reason was when upgrading, a majority percentage of replicas were "lost" in the routing change. Since without this, if a replicas hashed to the same host, it wouldn't try for the next slot like previously.

Also, from a availability standpoint, restarting or losing a host and having the the primary and replica(s) on the same host, could cause problems.

deniszh:
Hello @dctrwatson ,
Good idea, but it will break existing clusters, am I right?

dctrwatson:
Depends on when the cluster was created.
Created pre-routing change, this restores all the missing historical data.
Created post-routing change, then yes, it they will lose some/all their historical data.
(Both just applying to the replicas, not the primary)

deniszh:
Hi @dctrwatson ,
Then I better to make this thing optional in #452 
Please review!
I close that in favour of #452 . Thanks for your contribution!

PR #143, opened 2013-09-18
fix setup.py install_requires

tbarbugli:
Sadly enough (at least with pip install) package names are case sensitive

setup is trying to install twisted and txamqp instead of Twisted and 
txAMQP


obfuscurity:
Merge conflict resolved in #440.

PR #152, opened 2013-10-05
Stop making redundant deferred sendQueue calls

fhossain:
While running under a profiler I detected that we are creating a callback for every metric that is added to the queue. However, the queue is flushed in bulk. This leaves a lot of call backs with no data to flush. The optimization is to not queue a callback is one already exists.


obfuscurity:
@jssjr I'm ready to merge this in but I'd like to get your :+1:.

mleinart:
Did some testing of this over the weekend. Consistently got 20~25% increases in max throughput in the simple case relay routing to a single endpoint

pcn:
:checkered_flag:

PR #162, opened 2013-10-17
Added RELAY_METHOD remove-node-consistent-hashing

phufool:
This change adds a new relay method which determines the destination of the
metric by first removing a specified node from the metric. A case where this
is useful is when you want to aggregate metrics across all hosts, and the host
name is attached to the metric. This rule allows you to efficiently remove the host
name node without having to use any aggregation rules. It will then use
consistent hashing to group the same metrics to the same aggregator instance.

To use this relay method, add in the [relay] section:
# Removes first node of the metric

RELAY_METHOD = remove-node-consistent-hashing
REMOVE_NODE_INDEX = 0 

REMOVE_NODE_INDEX can have a negative indice.


EvanKuhn:
I need this exact functionality to properly run multiple aggregators on a single host.  Too bad this was never merged... :(

obfuscurity:
Is there something about rules-based relaying that won't address this use case?

briancervenka:
The specific use case here is this:
We send metrics in the form of: clustername.metricname.sum.nodename
Metrics are aggregated into clustername.metricname.sum_all.hosts
We send >10M metrics per minute
We run multiple aggregators, to divide load between multiple cores.

We need to make sure that all cluster metrics for a specific clustername.metricname.sum.* set are serviced by a single aggregator. Our aggregators then pass the individual metrics, along with some aggregated (sumall) metrics, down to the cache. This way, we can have an aggregate of the total cluster metrics, to prevent needing to recalculate those numbers in the graphite fronted. 

The rules based metrics don't seem to let us do hashing (ignoring the hostname) to distribute the metrics, we would need to make regexps for each specific metric and manually divide the metrics around. By removing the nodename before doing the hashing, we save a regexp lookup, consolidate use of the aggregators, and prevent needing to either have a relay rules table with all of our metrics, or a confusing relay rules file that tries to emulate hashing with creative regexps.

obfuscurity:
Just playing devil's advocate, but wouldn't it be better to design it as *inclusive* rather than *exclusive*, e.g. specify the nodes that you **want** to hash on, rather than the one(s) that you don't? I know this requires more information up front, but it feels more expressive and unequivocal to me.

briancervenka:
I can see use cases for each. However, in our case, the negative index (remove the second to last field) is most useful -- when I mentioned our stats are in the form of "clustername.metricname.hostname.sum", the metricname is sometimes an arbitrary number of fields, and all should generally be included in the hash, and anything that doesn't allow for removing the second to last field via negative indexing, would be harder to configure.

PR #163, opened 2013-10-17
Adds the ability to receive pickles over UDP

hufman:
One of the developers I work with was wondering why he couldn't send pickles over UDP, while he could send line data over UDP. This adds the ability, in case anyone else wants to do that.


obfuscurity:
@hufman So the intent here is not to add client UDP/pickle support (for inter-daemon communications), just the listener service for developers tooling their stuff accordingly?

obfuscurity:
@hufman After thinking about it a bit more, it seems like a benign addition -- but I worry about the confusion of **Yet Another Listener**, specifically one that few people will use. TBH I'm not even a fan of the UDP listener (I *want* deliverability guarantees for my metrics) but I can see why some might need it.

As it stands I'm going to close this out but if you can make a strong case for its inclusion I'll put in the work to merge the conflicts.

PR #176, opened 2013-11-05
fixed aggregator behavior on non-real time data points

plankthom:
occurs with clock differences, queued or replayed data
 * MAX_AGGREGATION_INTERVALS: consecutive number of extra aggregation
   buckets to keep, relative to most recent timestamp of previous data
   for that metric (and NOT system time)
 * MAX_AGGREGATION_AGE: maximum time (in number of aggregation intervals)
   to keep the buckets around (since system time of last update),
   defaults to 0 (no cleanup). Internally updated to
   MAX_AGGREGATION_INTERVALS if it is smaller that this setting.
 * do not write data that arrives to late (as it overwrites data that
   is stored), a warning is emitted to the aggregation log.


obfuscurity:
Thanks so much for this submission and my personal apologies for it languishing so long. I've moved this over to #439 to add a little bit of cleanup.

PR #178, opened 2013-11-05
Aggregation rules now search for exact regular expression pattern matches

feangulo:
We were experiencing an issue in which p99 values were being skewed by p999 values. We had the following aggregation rules:

production.all.response_time.p99 (10)   = max production.host.*.response_time.p99
production.all.response_time.p999 (10) = max production.host.*.response_time.p999

Our experience was that the aggregated p999 metric values matched the per host p999 metric values. However, the aggregated p99 metric values were considerably higher than the per host p99 metric values. In essence, the p99 and p999 metric values were being clubbed together.

With this code change, the aggregation rules will look for exact pattern matches.

- if it's p99: p99 does match, p999 doesn't match
- if it's p999: p99 doesn't match, p999 matches

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -379,3 +379,7 @@ MAX_AGGREGATION_INTERVALS = 5
 # In order to turn off logging of successful connections for the line
 # receiver, set this to False
 # LOG_LISTENER_CONN_SUCCESS = True
+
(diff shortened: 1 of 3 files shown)


obfuscurity:
This functionality was added in https://github.com/graphite-project/carbon/pull/120. Closing as a duplicate, but thanks! :sparkles:

PR #186, opened 2013-11-20
create an unsecure manhole if no key is provided

esc:
Make the manhole authentication configurable, easier to activate/deactive if
only config needs to be changed.


obfuscurity:
Support for this feature was recently merged in via #339. Thanks!

Issue #200, opened 2013-12-21
amqp.conf is missing AMQP_SPEC


zstyblik:
BRANCH: megacarbon

AMQP_SPEC is missing from amqp.conf and carbon-daemon, resp. AMQP consumer, will fail to start because of that. Now, this can be fixed by adding 'AMQP_SPEC=' into amqp.conf.
What I don't understand is why amqp code is way different when master and megacarbon branches are compared. I mean, looking at master branch, it seems to me like undefined configuration parameters/keys were handled better. In megacarbon, though, undefined  parameter/key is fatal error and daemon will bite a bullet in such case.

I guess I could try pull request for this one. I mean, it's a simple one-line add into conf file. :) But question is, shouldn't be amqp code more mature as in master branch?

'lib/carbon/service.py' in master:

```
amqp_spec     = settings.get("AMQP_SPEC", None)
```

and the same file in megacarbon:

```
amqp_spec = settings.AMQP_SPEC
```

PR #201, opened 2013-12-21
set AMQP defaults in conf.py rather than in-place

mleinart:
Noticed this after it was mentioned in #200
These conf settings have defaults specified in-place instead of in conf.py. This is a bit sloppy and also has the disadvantage that the types (string, int, et al) of the conf values aren't validated as they should be.

Moved these to conf.py where they belong


obfuscurity:
Looks like this was made obsolete by the changes in https://github.com/graphite-project/carbon/commit/88a646441cabf7def6721546c62ef19e5a1c7e1e#diff-75bda5fcd9a56b6bcbec28baefabe02c from #358. Please reopen if I'm mistaken.

PR #208, opened 2014-01-27
Better EL init scripts

johnseekins:
Improvements to the EL init scripts:
1) These init scripts try and look at whether a graphite process has open ports to determine whether it should be killed.
2) They can start, stop, etc individual processes as well as all specified processes
3) They are a little more portable in an EL environment as they don't assume defaults from graphite for a number of settings.


obfuscurity:
@johnseekins Sorry for the (ridiculously lengthy) delay in getting around to your PR. Unfortunately it looks like your patch pulled in a bunch of unrelated commits. I'm going to close this one down but if you'd like to resubmit a fresh PR with your initscript changes I promise a fast turnaround.

johnseekins:
No. That was a terrible commit. And I don't believe I'm using those init scripts any more, anyway...

Thanks for looking!

PR #216, opened 2014-02-19
Pluggable storage engines for carbon

posix4e:
Based on jbooth's  minimalist db plugin for carbon, (https://github.com/graphite-project/carbon/pull/210)

From jbooth:
" so that it can record to other storage engines besides whisper, just as graphite-web can pull up other finders besides the StandardFinder. It has 5 methods:

info
setAggregationMethod
create
update_many
exists

and is configured by a single config variable, DB_INIT_FUNC, which should be a visible function that yields an object fulfilling the above contract.

I've got an example plugin for Hbase at https://github.com/posix4e/graphite-data --  -- we could add other storage engines to that repo and I'd be happy to contribute the whole repo over to graphite-project as well. "

I rebased it to a single patch and will track master.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -29,6 +29,8 @@
 #   PID_DIR        = /var/run/
 #
 #LOCAL_DATA_DIR = /opt/graphite/storage/whisper/
+# initializes graphitedata plugin DB
+# DB_INIT_FUNC=graphitedata.plugin.WhisperDB
 
 # Specify the user to drop privileges to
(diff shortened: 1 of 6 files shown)


obfuscurity:
cc @woodsaj

iksaif:
I think this is obsolete, we now have "Database" plugins.

obfuscurity:
Agreed, the `TimeSeriesDatabase` class serves as the pluggable model for Carbon.

PR #227, opened 2014-03-11
add zlib support.

dnozay:
I like the hassle-free UDP transport for sending packets,
and I like zlib for sending a lot of data in smaller packets.


obfuscurity:
I'm torn on this patch. In one case, it sounds like a neat feature add. OTOH, I can't find a single instance where a user has requested compression. I'm tempted to rebase and resubmit this, but I'm a little concerned that you deemed the change worthy enough to add yourself to the Copyrights.

/cc @drawks @SEJeff @jssjr @mleinart for opinions

jssjr:
This looks like a useful way to reduce packet-per-second on the receiver at the expense of higher CPU load due to decompression overhead. I can see situations where this might be worthwhile.

The copyright modification isn't needed and should be removed. Though, it would be nice to have an AUTHORS file.

dnozay:
should the milestone be updated to 1.2.0?

piotr1212:
This is an old one. Needs to be rebased and made compatible with py3.

PR #231, opened 2014-03-21
Added the option to specify the minimum number of datapoints per update operation

pratX:
Motivation:
1.) Avoid a large a number of small disk-writes.
2.) A more predictable write pattern across multiple carbon-cache instances i.e. every carbon-cache instance getting a share of disk I/O.

Modifications to meet the above points:
Modified writer.py to make carbon-cache do a whisper.update_many() only for those metrics which have more than MIN_DATAPOINTS_PER_UPDATE datapoints in cache. And this constraint is removed when the cache size grows above LOW_CACHE_SIZE, and comes back into effect when cache size shrinks below LOW_CACHE_SIZE.
LOW_CACHE_SIZE is set to 80% of MAX_CACHE_SIZE by default.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -63,6 +63,17 @@ MAX_UPDATES_PER_SECOND = 500
 # the files quickly but at the risk of slowing I/O down considerably for a while.
 MAX_CREATES_PER_MINUTE = 50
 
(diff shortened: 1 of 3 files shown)


obfuscurity:
Closing this PR in favor of #437.

PR #233, opened 2014-03-24
Adds logging about why an input line is invalid

hufman:


withnale:
How about making this not log if the line length is zero? We use tcp-open healthchecks on our load balancer and have a local patch for disabling huge numbers of these entries...

SEJeff:
@withnale: care to send a PR?

toni-moreno:
+1:
This will be very usefull to debug metrics !!

obfuscurity:
I would :heart: to see tests for this.

obfuscurity:
Conflict resolved and merged in #436. Thanks!

PR #239, opened 2014-05-13
Enhanced rebased #140

pabigot:
Merge request #140 (add a setting to inhibit diagnostics when forwarding metrics) no longer applied cleanly; the first commit updates it.

The second commit corrects the test so the diagnostic is provided only when the original metric did not contribute to an aggregation, and not in the much more common case when it was not clobbered by an aggregate metric.


obfuscurity:
Fixed in #435.

PR #241, opened 2014-05-21
Fix crashes due to auto-type in Python

zstyblik:
These two commits do fix crashes caused by auto-type feature in Python. We expect instance of one type, however Python automatically creates instance of int/str which doesn't have or isn't capable of wanted operations.

### DIFF
--- a/lib/carbon/instrumentation.py
+++ b/lib/carbon/instrumentation.py
@@ -122,13 +122,13 @@ class InstrumentationService(Service):
           events.metricGenerated(metric_path, datapoint)
 
       else: # counters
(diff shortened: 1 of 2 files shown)


obfuscurity:
Sorry, I have to agree with @drawks here. If there's a problem with https://github.com/graphite-project/carbon/blob/megacarbon/bin/ceres-maintenance we should try to fix it at the source. Note that Ceres support has not yet been merged over to master from the megacarbon branch yet. /cc @mleinart

SEJeff:
Absolutely the right call.

PR #248, opened 2014-06-18
Use the RedHat init scripts on CentOS, too

Al2Klimov:
Because CentOS is based on RHEL, it makes sense to use the RedHat init scripts on CentOS, too.


obfuscurity:
LGTM, thanks!

PR #258, opened 2014-07-21
Ensure "except"s only catch "Exception"s (master)

steve-dave:
Bare except blocks catch SystemExit and were making a significant number of our attempts to stop carbon daemons fail.

Courtesy @Roguelazer with additions


steve-dave:
Covered by https://github.com/graphite-project/carbon/commit/00b5b881b35b150a9552712b8931eb5426c1f115 and https://github.com/graphite-project/carbon/commit/c6cd6c9245f93dcc65f9fea60c6b28a78c8bdff8

Issue #167, opened 2013-10-21
carbon-cache cpu usage jumps to 150% when setting MAX_CACHE_SIZE != inf


mabrek:
I've set 
MAX_CACHE_SIZE=1000000
USE_FLOW_CONTROL=False
and noticed jump in carbon-cache cpu usage (150%)
When I reverted it back to default 
MAX_CACHE_SIZE=inf
USE_FLOW_CONTROL=true
cpu usage dropped to 15-30%

version 0.9.12

PR #266, opened 2014-08-14
issue #167: Don't iterate on the cache for every call to size

nukemberg:
As mentioned in [issue 167](https://github.com/graphite-project/carbon/issues/167), calculating MetricCache size using reduce creates a large number of futex calls causing high cpu usage. This patch keeps track of MetricCache size using an auxiliary variable. Since data is pushed and popped from different threads size tracking is not accurate, but for the purpose of limiting memory usage i feel this is good enough.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -23,6 +23,7 @@ except:
 
 class _MetricCache(defaultdict):
   def __init__(self, defaultfactory=deque, method="sorted"):
+    self.size = 0
     self.method = method
     if self.method == "sorted":
       self.queue = self.gen_queue()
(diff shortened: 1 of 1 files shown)


loopping:
I think I fixed the size tracking for "sorted" method.
Please note I'm not a developer and don't fully understand the way carbon works :) 
In cache.py,  function pop reads like this:
<pre><code>
  def pop(self, metric=None):
    if not self:
      raise KeyError(metric)
    elif not metric and self.method == "max":
      metric = max(self.items(), key=lambda x: len(x[1]))[0]
      datapoints = (metric, super(_MetricCache, self).pop(metric))
    elif not metric and self.method == "naive":
      datapoints = self.popitem()
    elif not metric and self.method == "sorted":
      metric = self.queue.next()
      # Save only last value for each timestamp
      popped = super(_MetricCache, self).pop(metric)
      self.size -= len(popped)
      ordered = sorted(dict(popped).items(), key=lambda x: x[0])
      datapoints = (metric, deque(ordered))
    if not metric.method == "sorted":
      self.size -= len(datapoints[1])
    return datapoints
</code></pre>
When I traced, I could notice that the size of "popped" was different of "datapoints".
Now, the size returns to 0 when no more items are in cache.

PR #323, opened 2014-11-04
twistd syslog support

rppala90:
Unable to use twistd option syslog when starting carbon cache from /usr/bin/carbon-cache.


obfuscurity:
Thanks!

Issue #185, opened 2013-11-19
memory leak in relay and aggregator daemons


dkulikovsky:
I've got a leaking megacarbon instances. The load is low - 50-100k metrics per minute. Config is simple - i have a pipeline 2 relays with aggregated-consistent-hashing -> 1 aggregator with consistent hashing -> 2 writers. 
Instances that run relay or aggregation role consumed tons of memory:

```
top - 16:03:22 up 6 days,  1:09,  7 users,  load average: 143.34, 76.86, 48.83
Tasks: 414 total,   4 running, 410 sleeping,   0 stopped,   0 zombie
Cpu(s):  8.2%us,  0.5%sy,  0.0%ni, 90.0%id,  1.2%wa,  0.0%hi,  0.1%si,  0.0%st
Mem:  65949272k total, 65443728k used,   505544k free,  1215012k buffers
Swap:        0k total,        0k used,        0k free,   269768k cached

  PID USER      PR  NI  VIRT  RES  SHR S %CPU %MEM    TIME+  WCHAN     COMMAND                                                                                                                        
 2824 www-data  20   0 23.7g  23g 1816 R 99.4 37.5  34:38.02 -         /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-relay-st01 start                                                  
 2897 www-data  20   0 11.9g  11g 1760 R 99.4 18.8   9:28.29 -         /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-relay-st02 start                                                  
15538 www-data  20   0 23.7g  23g 1796 S 44.7 37.5  14:47.56 ep_poll   /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-aggregator-st01 start                                             
 1125 www-data  20   0  299m 121m 1860 S  7.9  0.2  31:58.85 poll_sche /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-writer-st02 start                                                 
 1120 www-data  20   0  310m 146m 1860 S  7.4  0.2  37:09.29 poll_sche /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-writer-st01 start                                                 
 1118 www-data  20   0  310m 146m 1860 S  3.2  0.2  20:11.00 ep_poll   /usr/bin/python /var/lib/graphite/bin/carbon-daemon.py carbon-writer-st01 start                                                 
[...]

PR #335, opened 2014-12-02
add LRU cache and max size setting to relay to fix memory leak

bruce-lyft:
fix for https://github.com/graphite-project/carbon/issues/185

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -304,6 +304,9 @@ MIN_RESET_RATIO=0.9
 # reset connections for no good reason.
 MIN_RESET_INTERVAL=121
 
+# cache size for aggregation metrics.  Helps control memory use of relays.
+AGGREGATION_CACHE_SIZE=10000
+
 [aggregator]
 LINE_RECEIVER_INTERFACE = 0.0.0.0
 LINE_RECEIVER_PORT = 2023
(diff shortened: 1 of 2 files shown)


mleinart:
This fix looks like it'll work to prevent the cache from growing unbounded, however I'm not sure it's practical to limit the size in this way.

Think about the typical metrics flow - generally each actively updating metric will be coming in at a regular interval, e.g. every 1 minute. If we have a limit of say 1000 items in this cache, only the last 1000 metrics seen in a given minute will be cached for the next inflow of metrics. If we assume the metrics come in roughly the same order, those 1000 will be evicted before they ever see a cache hit.

FWIW, I've committed be2e4f0d91 for now as suggested in #185 which should help in many cases, though I think it's still an open question as to whether it's a net gain caching these to avoid regex operations

bruce-lyft:
@mleinart  The problem with the fix in be2e4f0 is that memory use is still unbounded.   In our case, we have new metrics coming in on a regular basis, with a short time horizon (collectd/statsd metrics that are per-host, for ephemeral hosts, but which are aggregated; for example, CPU average for a group of hosts).   Without this fix, we see carbon-aggregator memory use grow without bound and eventually crash.

obfuscurity:
@bruce-lyft I think the aggregator leak in #185 is almost certainly the result of feeding aggregated metrics back into the aggregator loop as described in #455 and #560. There's a proposed fix in #439 but it needs some TLC before we can merge it. There's also an external fix applied to https://github.com/criteo-forks/carbon/commit/825693c97e57c7cdc36d0b0d758694010ce2be53 but I'd like to get feedback from others before we consider either of these.

/cc @mleinart @drawks @iain-buclaw-sociomantic

obfuscurity:
Deleting my previous comment and moving it to #185.

obfuscurity:
Closing as discussed in https://github.com/graphite-project/carbon/issues/185#issuecomment-246122400.

PR #338, opened 2014-12-05
Adds direct AMQP exchange option in addition to the default topic one

milosgajdos83:
With this PR you have an option to run multiple carbon-relays which can
essentially behave as "workers". These carbon "workers" are reading in
metrics from the ```direct``` AMQP exchange and distribute them to carbon-caches
or to other relays based on relay method - there is no binding key necessary.
Previously you could only distribute based on binding/routing key
so you were not able to have true relay WORKERS because having them
running like that would generate duplicate writes on your drive.
With topic exchange and default binding key '#' you'd always get duplicate writes.
Direct exchange allows metric to be consumed by ONE carbon cache worker - the
mirroring is then handled vi carbon itself by specifying a particular carbon configuration.


obfuscurity:
@milosgajdos83 Would you mind rebasing this?

milosgajdos83:
I tried, but from the looks of it `lib/carbon/service.py` has changed significantly since my PR :-)

I need to spend a bit more time with this to figure out how to get this synced without breaking anything :-/

obfuscurity:
I can probably rebase it for you but I'll need you (or someone that uses AMQP) to test that it still works as designed.

milosgajdos83:
I'm happy to retest it once it's rebased.

obfuscurity:
@milosgajdos83 Moving to #445, closing this one.

PR #344, opened 2014-12-19
Adding init.d scripts for Debian

gfaivre:
Conversion of the Redhat based init.d scripts to the Debian way.


obfuscurity:
Looks good. I think we can also :bomb: the old postinstall script at https://github.com/graphite-project/carbon/blob/master/debian/carbon.postinst. /cc @bmhatfield

PR #352, opened 2015-01-05
Remove half-baked documentation in github README

ryan-williams:
Point to readthedocs.org instead, which is currently the most complete documentation source afaict

PR #353, opened 2015-01-06
Simplify README and add diagram

obfuscurity:
Feeling motivated by @ryan-williams to fix our READMEs and point to the official documentation. Supercedes #352.


deniszh:
@obfuscurity, do we need to backport it to 0.9.x btw ? I can do that.

PR #354, opened 2015-01-06
Simplify README and add diagram

deniszh:
Backport #353 to 0.9.x
N.B. - maybe it's need to be pointed to https://graphite.readthedocs.org/en/0.9.13/ instead of https://graphite.readthedocs.org/en/latest/ but 0.9.13 is not exist yet and pointing to 0.9.13-pre1 is quite pointless IMO...


obfuscurity:
Unnecessary. I doubt anyone will read this. The change to master was only for folks that land on the project and need to be directed to the real docs.

PR #358, opened 2015-01-10
Port Processor/Pipeline refactors from megacarbon

mleinart:
The 'megacarbon' branch contains a large refactor that implements the transformation (aggregation, rewriting) and output (cache/writer or relay) as a pipeline of processors. This change separates out that refactor from the other changes in the branch without any of the config compatibility changes that it originally was paired with. I attempted to make this change as small and contained as possible without any unnecessary changes (style, etc) mixed in. Even so, it's pretty large. I expect this'll be the biggest changeset coming from megacarbon. Following ones should be much more contained.

#### Motivation ####
I've merged megacarbon with master to prepare for a merge several times, but each time have been asked to hold off merging because of the big scary config refactor it includes. The megacarbon branch is also a mess in terms of git history as it was worked on as a complete fork by chrismd without a coherent commit history.  For us to be able to take advantage of some of the great refactorings in the branch as well as Ceres write support, I've started the process of manually pulling out refactors and implementing unit tests along the way.

This refactor, in addition to allowing further code to be pulled in from megacarbon, sets us up to be able to start work a few interesting things including the ability to load arbitrary site-specific metric processing plugins. It will also allow us to start to figure out how we might allow construction of arbitrary pipelines as megacarbon allowed (say I want to rewrite metric names in a relay - today that requires two processes, a carbon-relay and a carbon-aggregator)

#### Changes ####
This introduces a new interface for a Processor: https://github.com/mleinart/carbon/blob/88a646441cabf7def6721546c62ef19e5a1c7e1e/lib/carbon/pipeline.py
The flow through Carbon is implemented as a pipeline from a Receiver (e.g. MetricLineReceiver), through an arbitrary processing pipeline, ending at a writer (carbon-cache) or relay (carbon-aggregator, carbon-relay). 

Changes to aggregator, rewrite, and client are straightforward and ported directly from megacarbon with minimal modification.

Along the way in cache.py, I unfortunately ended up deciding to pick up most of ead2176f05 as that had been merged to megacarbon but NOT master and I was feeling too lazy to spend the many hours manually merging the many many conflicts from that merge (ugh, if only we squash-committed all merges). The changes here are nearly directly copied from the already-merged megacarbon version with megacarbon's instrumentation refactors and config rename removed.

writer.py similarly has changes from ead2176f05 missing so it's copied directly from 0.9.x. I didn't write any tests for this as it'll go through a great deal of changes when Ceres support is merged in.

The startup code in service.py is a mix between megacarbon and the existing master branch as megacarbon's startup completely changed with the move from multiple carbon daemon types to a single universal daemon. The version here effectively hardcodes the pipelines to match the existing carbon-relay, carbon-aggregator, and carbon-cache daemons. In the future it should be easy to make these pipelines configurable in some fashion.

#### Testing ####
I've written unit tests for most of what I touched, excluding writer.py as noted above. I've also done as much manual testing as I can think of to validate it. The tests are a start, but coverage is nowhere near 100%


SEJeff:
Send me your addy privately. I owe you a nice bottle of whiskey for doing this.

drawks:
:astonished: :+1:

obfuscurity:
So :fire:.

SEJeff:
Question about this @mleinart. Does this mean that the magic with Ceres where a metric can be split up over multiple caches will be in master?

Please merge!

obfuscurity:
Ceres

mleinart:
@SEJeff yup I can work on getting ceres in with this part done

toni-moreno:
Hi @obfuscurity , @mleinart .

After this merge, is current master ready to update from  a previous commit (https://github.com/graphite-project/carbon/commit/dec9dfed89ea19e756fd5a45ff6462874b03b904) a currently running graphite production server ?

Is backwards compatible with my current carbon.conf ? 
Can I stop/start with my current /etc/init.d/carbon-cache.sh / /etc/init.d/carbon-relay.sh ?
Can Increase the resource needs (cpu/memory)  after this merge?

PR #359, opened 2015-01-12
Restore correct default for RELAY_METHOD

mleinart:
01062f182 inadvertently changed RELAY_METHOD's default to a new value for relay-rules
This reverts that part of the commit to restore backward compatibility


lib/carbon/service.py
```diff
@@ -203,7 +203,7 @@ def setupRelayProcessor(root_service, settings):
     aggregation_rules_path = settings["aggregation-rules"]
     RuleManager.read_from(aggregation_rules_path)
     router = AggregatedConsistentHashingRouter(RuleManager, settings.REPLICATION_FACTOR)
-  elif settings.RELAY_METHOD == 'relay-rules':
+  elif settings.RELAY_METHOD == 'rules':
     router = RelayRulesRouter(settings["relay-rules"])
```
obfuscurity: Is this correct?

marco-hoyer:
I think this is correct. Since there is also a check in conf.py to ensure RELAY_METHOD is rules, consistent-hashing or aggregated-consistent-hashing the relay-rules setting would work either.
Should be merged as soon as possible since master is broken atm.

```python
if settings["RELAY_METHOD"] not in ("rules", "consistent-hashing", "aggregated-consistent-hashing"):
```

marco-hoyer:
thx

Issue #333, opened 2014-11-28
REPLICATION_FACTOR  send metrics over carbon-caches on the same machine.


toni-moreno:
We are testing the following graphite scheme, Our goal is to have all data on graphite1 replicated ( as a backup server in graphite2) . After some test,  we can see some metrics only on one machine, ( we are using carbon-lookup tool from carbonate also to check where it is sending metrics the relay) , after some test We have seen that REPLICATION_FACTOR + consistent_hashing doesn't know anything about hosts  machines only carbon destinations. This is , the relay is sending the same metric twice over the same machine with the following results.

* no data backup
* overload of the disks.

how can configure this scheme avoiding this result?

![image](https://cloud.githubusercontent.com/assets/5883405/5225547/7bb0b0d6-76e5-11e4-8609-ee498798a50e.png)

toni-moreno:
Finally using carbon-c-relay as a proxy to other carbon servers.

Thank you very much @deniszh.

PR #365, opened 2015-01-20
add MIRRORS

andyxning:
Recently, my team came into a problem with  carbon-relay. Here it is.
We use carbon-relay with RELAY_METHOD be consistent-hashing and REPLICATION_FACTOR be 2 with 2 backends with carbon-cahce. However, as time goes, the data is growing more and more and we finally make a decision to add more machines and use consistent-hashing to do load balance between them. At the same time we also want to get a full dumped metric info to do some other work. 

I read the source code about carbon 0.9.x and some documents and finally find out that i can not get a solution with current carbon's functionality. So, i finally make this pull requests to add  a new configuration option in carbon-relay. 

MIRRORS
-----
##### Description
* This defines mirror destinations for relay to dump a full metric info no matter how you configure RELAY_METHOD. This can be helpful when you want to duplicate all metric info as well as use RELAY_METHOD, like consistent-hashing, to do some load balance or else. 
* Note this option take the same argument format as DESTINATIONS
* This option uses TCP + pickle to sent data
* By default, this option will be empty and has no influence on default relay behaviors.  

##### Configuration
* MIRRORS = 127.0.0.1:2004, 127.0.0.1:2005  

#####  Usages
assume below:
```
RELAY_METHOD = consistent-hashing
MIRRORS = 127.0.0.1:2004
```
with settings above, we can get a fully dumped metric info to MIRRORS destinations with also the behavior of consistent-hashing to do some load balance.

NOTIFICATION
-----
after the PR i have commit another one, with a bug fix about MIRRORS. So, you should use both commis to make a merge.


toni-moreno:
hi @andyxning , can this PR remove one relay tier in this situation? ( https://github.com/graphite-project/carbon/issues/333) 

would  I be able do the following layout with this PR ?

image

andyxning:
​i see [#333](​https://github.com/graphite-project/carbon/issues/333). I
think this MIRRORS can meet your requests.

I am not sure i have understood what you said in [#333](
https://github.com/graphite-project/carbon/issues/333), so if i am wrong
please let me know.

I think you want to make load balance with carbon-relay and also want to
make a fully dumped info to make a backup. So, if it true, this MIRRORS
option will be definitely suit you.

And thus you can just use one carbon-relay instead of four relays, as
dist-01a, dist-01b, dist-01c, dist-01d] with consitent-hashing and can have
MIRRORS points to your backup server.

MIRRORS will not impact the normal behavior about RELAY_METHOD, instead it
just use the already existing code about receiving and sending metric of
carbon-relay. So it will work strongly just like carbon-relay.




++++++
Ning Xie


2015-01-20 16:02 GMT+08:00 Toni Moreno <notifications@github.com>:

toni-moreno:
I'm still thinking on the layout we need and I think it is not possible if you only do a complete metric replication .

condider the most easy layout.

image

If I configure mirror to send data to each other metrics ( as in the next picture) could itself enter in a infinite loop, isn't it?

image

obfuscurity:
You say that you "can not get a solution with current carbon's functionality". Can you elaborate on this claim?

obfuscurity:
As a point of reference, we had a full metrics feed replicated among two DCs, as described [here](https://gist.github.com/obfuscurity/63399584ea4d95f921e4).

toni-moreno:
hi @obfuscurity I  already have  the first layout  in the previous post working ok. But  I would like to have a single carbon-relay doing the 2 things itself (if possible) to eliminate a  relay tier

obfuscurity:
@toni-moreno That was actually directed at @andyxning. But are you saying that #333 can be closed?

andyxning:
I mean that in our situation which use REALY_METHOD being
consistent-hashing to do some load balance and in which we also want an
option to config a destination to dump a full info singly with out the
influence of consistent-hashing DESTINATIONS.

we formally use consistent-hashing with replication-factor be 2 and two
backends, thus all  the backends have the same data. Later, we split fully
dumped info by adding a destination in the DESTINATIONS option with
 replication-factor be 3. However, with data growing, now we want to add
two backends  also with the consistent-hashing method and
replication-factor be 2.

Thus the question comes, we can not still put the dumped destination in
DESTINATIONS option, because it will not get a fully dumped info anymore
only part of it. We still not want to add more layers into our system,
because it may make our system more complicated and thus more work to make
it correctly.

Under the above situation, i read the carbon source code of 0.9.x and
finally add the MIRRORS option.



++++++
Ning Xie


2015-01-20 23:29 GMT+08:00 Jason Dixon <notifications@github.com>:

obfuscurity:
I don't think that "not wanting to add another layer" is a valid reason for introducing this change. I still don't understand how this is substantively different than REPLICATION_FACTOR=2 with consistent-hashing. Can someone else explain to me what I'm missing?

andyxning:
yes, that's absolutely right. If you do configure mirror to send data to
each other metrics, it will enter in a infinite loop.

I put the above answer and add MIRRORS option under the assumption that the
fully dumped info will be used in other subsystems, not anymore in carbon.

Sorry for not making my idea clearly.



++++++
Ning Xie


2015-01-20 22:24 GMT+08:00 Toni Moreno <notifications@github.com>:

andyxning:
Sorry for my poor english. Here is a example.

Suppose we use consistent-hashing method and replication-factor be 2 with
backends, backend1 and backend2.  now with the consistent-hashing method
backend1 and backend2 will both have the same data.

now, we add  a new destination, dump-backend, to the DESTINATIONS also with
consistent-hashing method and replication-factor be 3. Now, DESTINATIONS
will be ​backend1, backend2 and dump-backend and all the three destinations
will also have the same data cause the replication-factor be 3.

After that, now we want to add two new backends, backend3 and backend4, to
DESTINATIONS with replication-factor be 2 and also consistent-hashing
method. We do so just want to do some load balance.

what i want is that carbon-relay will do load balance with
consistent-hashing in backend1, backend2 , backend3 and backend4 and also
config dump-backend to get a full dumped data.

If i config dump-backend with backend[1-4] in DESTINATIONS, thus
consistent-hashing will have five machines to do load balance not in four
backend machines. And so, i can not get a fully dumped info only parts
actually.

we can not find an existing option can do this job. So, my team come into
that we can add a new config to do the dump data job with no influence on
the existing relay methods DESTINATIONS.



++++++
Ning Xie


2015-01-21 10:19 GMT+08:00 Jason Dixon <notifications@github.com>:

obfuscurity:
@andyxning I appreciate your effort to explain your setup but I feel like I must still be missing something. Do you have a simple diagram that illustrates your configuration?

andyxning:
​As shown in below graph.

G1 is the original arch we use.

>>Suppose we use consistent-hashing method and replication-factor be 2 with
backends, >>backend1 and backend2.  now with the consistent-hashing method
backend1 and backend2 >>will both have the same data.

G2 is the ​one that we add a new backend, dump-backend.

>>now, we add  a new destination, dump-backend, to the DESTINATIONS also
with consistent->>hashing method and replication-factor be 3. Now,
DESTINATIONS will be ​backend1, >>backend2 and dump-backend and all the
three destinations will also have the same data >>cause the
replication-factor be 3.

Until now, everything is ok. Thus we come into the graph G3. G3 is the arch
that we thought with the requirement below:

>>After that, we want to add two new backends, backend3 and backend4, to
DESTINATIONS >>with replication-factor be 2 and also consistent-hashing
method. We do so just want to do >>some load balance.

>>what i want is that carbon-relay will do load balance with
consistent-hashing in backend1, >>backend2 , backend3 and backend4 and also
config dump-backend to get a full dumped data.

as is shown in G3, If i config dump-backend with backend[1-4] in
DESTINATIONS, thus consistent-hashing will have five machines to do load
balance not in four backend machines. And so, i can not get a fully dumped
info only parts actually.

Thus in G3 we will get "a.b.c" in backend1 and backend2 and "d.e.f" in
backend3 and dump-backend. This is wrong because we want dump-backend get
"a.b.d" and "d.e.f" and backend1 and backend2 get "a.b.c" and backend3 and
backend4 get "d.e.f".






++++++
Ning Xie


2015-01-21 12:02 GMT+08:00 Jason Dixon <notifications@github.com>:

andyxning:
sorry for last email without the mentioned graph. Here it is.



++++++
Ning Xie


2015-01-21 14:28 GMT+08:00 Andy Xie <andy.xning@gmail.com>:

obfuscurity:
So if I understand you correctly now, you're adding `MIRRORS` just to avoid running an extra layer of relay(s) to perform replication between your two backend clusters. If that's correct, then I'm :-1: on merging this.

You can get the same effect by just running another carbon-relay on the same host. You don't need separate hardware, and you don't need `MIRRORS` to accomplish this. You just need another carbon-relay that handles replication at the first "layer", then another relay to handle load-balancing at the second "layer". Each relay has their own `DESTINATIONS` and `REPLICATION_FACTOR`.

I mentioned this type of setup in my [previous comment](https://github.com/graphite-project/carbon/pull/365#issuecomment-70671880).

andyxning:
Yes, that's right.

However, i still think if we can do the two things in one instance without
adding any influence to the carbon-relay's original functionality, it may
be not that bad to do this than adding more carbon-relay instance in the
processing chain. The latter one add more layers in the processing chain
and will make the entire arch more fragile and we must pay more attention
to operate this, like more configuration files and we still need to monitor
two instances at  the same time.



++++++
Ning Xie


2015-01-21 22:49 GMT+08:00 Jason Dixon <notifications@github.com>:

obfuscurity:
I'm strongly against this. It may seem like "not a big deal" to you, but you're effectively duplicating *in code* what can already be accomplished *via configuration*. This adds unnecessary bloat to the code base and could easily confuse other users. Look how long we've been discussing it and **I actually understand how carbon-relay works**. :stuck_out_tongue: 

None of the other project owners or committers have stepped in to offer their opinion, so I'll leave this open for a while. But I'm :-1: on this change.

andyxning:
Since it won't be accept i will close this PR, however i will retain this feature in my forked version.

TheLinuxNinja:
While for a new person this would seem to be a logical addition, I agree that adding MIRRORS is a bad idea. I already have this functionality in our setup using a separate relay tier. Actually, I have 5 tiers in each of 4 datacenters, all replicating to each other.

andyxning:
​I still think if we can do some thing in one layer, and at the same time
nothing complicated thing is imported to accomplish this, then we should at
least consider about this.

In your condition, 5 tiers is configured in each datacenters, if it is just
for what MIRRORS option does then i would think that i will heavily burden
the OPS to maintain all 5 tiers to work correctly to make sure that it is
do the right thing.

i think we should make arch more flat if possible, isn't it? [?]



++++++
Ning Xie


2015-01-29 0:47 GMT+08:00 Bryan L. Gay <notifications@github.com>:

obfuscurity:
You make it sound as if we're being stubborn about this, but I don't think we are. I *have* considered it and I'm personally against it, since the functionality already exists. To me adding this would be an anti-pattern and would potentially confuse users.

If you convince one of the other maintainers that this is a good thing, I won't fight its inclusion, but I'm not going to merge it myself.

PR #367, opened 2015-01-21
Close connection after n seconds if no data is received

piontas:
It allows carbon to close connection after n seconds if no data is received from it. The client will then have to reconnect in order to continue sending data. It's fixed using helper in Twisted called: "TimeoutMixin".

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -30,6 +30,9 @@
 #
 #LOCAL_DATA_DIR = /opt/graphite/storage/whisper/
 
+#Twisted TCP server timeout if no read after n seconds
+#TIMEOUT = 60
+
 # Enable daily log rotation. If disabled, a new file will be opened whenever the log file path no
 # longer exists (i.e. it is removed or renamed)
 ENABLE_LOGROTATION = True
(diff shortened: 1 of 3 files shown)


piontas:
Yes, thats true, but we have a lot of issues that carbon opened a lot of connections, didn't close any and just stopped working. I can set default as None as You said and it will be an option to configure.

pu239ppy:
Any chance this PR can get merged soon?
This would be extremely helpful to us as well, we point all logs to a VIP managed by keeplived, when keepalived state changes the existing connections never go away.  At some point carbon-relay runs out of open files, at that point not only does not not accept connections, it proceeds to log about it at a very high rate filling up disk.

obfuscurity:
I think it needs some minor additional documentation as @mleinart mentions. I'd also prefer to see a more meaningful name, e.g. `CONNECTION_TIMEOUT`. Lastly, it needs to be rebased.

pu239ppy:
@piontas Marek, are you still interested in this PR, would you mind doing the work needed? per @obfuscurity it would seem that all you need is a configuration file comment explaining the caveats and also renaming the variable to something more meaningful such as METRIC_CLIENT_IDLE_TIMEOUT

piontas:
Yes,
I will fix it tomorrow.

pu239ppy:
Would really love to see this merged, pardon my impatience, I cherry-picked the commits and made sure they can be merged to current master in my PR #429 
Would someone mind having a quick look at it?

obfuscurity:
Merged #429, closing this one.

PR #377, opened 2015-02-04
Get rid of a few more bare "except" calls (master)

steve-dave:
There were still a few hanging around.


steve-dave:
@sejeff

steve-dave:
@obfuscurity

PR #397, opened 2015-04-16
Could we get this in 0.9.x? Added the offending line to the error output, when otherwise skipping a ...

piotr1212:
this is from 00422a1e6b24a73c99ad20adfe4166c8f709c12a 
Could we get it in 0.9.x?

...mangled Metric line.

Conflicts:
	lib/carbon/protocols.py


piotr1212:
ping

obfuscurity:
Thanks!

Issue #383, opened 2015-02-13
carbon-relay stuck in AlreadyCalledError


lambdafu:
Hi,

I just updated my stack from an Oct 2013 version to current versions, and am getting a ton of AlreadyCalledError failures in carbon-relay.  Once it gets in these, it seems to be stuck (retrying every couple of seconds with the same problem), until it is restarted or dies because of other bugs (I was using the pypi 0.9.13 "release" and thus saw #364 also).

Somewhat at a loss how to get more information, as I don't know twisted or carbon internals, and it seems related to asynchronous operation.  Help! :)

My overall setup is a single carbon-cache (with pickle interface) and one carbon-relay for each protected namespace (customer, with line interface).  The relay's only job is whitelisting.

I am using the ubuntu latest (14.04) docker image with its twisted and other packages, and graphite from pypi (some of the dependencies are for mozilla's circus).

From the relay error log:
```
08/02/2015 19:20:47 :: [console] Starting factory CarbonClientFactory(127.0.0.1:2004:None)
08/02/2015 19:20:47 :: [clients] CarbonClientFactory(127.0.0.1:2004:None)::startedConnecting (127.0.0.1:2004)
08/02/2015 19:20:47 :: [clients] CarbonClientProtocol(127.0.0.1:2004:None)::connectionMade
08/02/2015 19:20:47 :: [console] Unhandled Error
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/twisted/python/log.py", line 88, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/lib/python2.7/dist-packages/twisted/python/log.py", line 73, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/lib/python2.7/dist-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/lib/python2.7/dist-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/usr/lib/python2.7/dist-packages/twisted/internet/posixbase.py", line 619, in _doReadOrWrite
    why = selectable.doWrite()
  File "/usr/lib/python2.7/dist-packages/twisted/internet/tcp.py", line 593, in doConnect
    self._connectDone()
[...]

PR #399, opened 2015-04-17
0.9.x fix stuck carbon relay

mleinart:
Addresses #383 

There appears to be a race that happens when the queue is full and several destinations reconnect at the same time, all send points along, and all try and signal that the queue's no longer full. One of these triggers the callback, but another hits the callback before the first one is complete.

This is a quick check to the callback state to ensure it's not called twice before being reset in this case


mleinart:
wtf

mleinart:
Derp, merge to wrong branch

PR #400, opened 2015-04-17
Fix race condition in carbon client queue full signalling

mleinart:
Addresses #383

There appears to be a race that happens when the queue is full and several destinations reconnect at the same time, all send points along, and all try and signal that the queue's no longer full. One of these triggers the callback, but another hits the callback before the first one is complete.

This is a quick check to the callback state to ensure it's not called twice before being reset in this case


mleinart:
This should cherry-pick cleanly into master as wel

PR #402, opened 2015-04-21
Remove log.debug in cache and writer

### LINKED ISSUE #401: log.debug has no effect
There are a handful of places where we call `log.debug()` but this doesn't seem to be supported by our logging module (inherited from Twisted?). I propose that we switch these to `log.msg()` but I'd like to hear from others.

#### 0.9.x
```
$ grep -r log.debug *
lib/carbon/cache.py:        log.debug("Sorted %d cache queues in %.6f seconds" % (len(queue), time.time() - t))
lib/carbon/client.py:    log.debug("Couldn't import signal module")
lib/carbon/client.py:      log.debug("Installing SIG_IGN for SIGHUP")
lib/carbon/writer.py:    log.debug("Couldn't import signal module")
lib/carbon/writer.py:          log.debug("Installing SIG_IGN for SIGHUP")
```
#### master
```
$ grep -r log.debug *
lib/carbon/cache.py:        log.debug("Sorted %d cache queues in %.6f seconds" % (len(metric_counts), time.time() - t))
lib/carbon/client.py:    log.debug("Couldn't import signal module")
lib/carbon/client.py:      log.debug("Installing SIG_IGN for SIGHUP")
lib/carbon/writer.py:    log.debug("Couldn't import signal module")
lib/carbon/writer.py:          log.debug("Installing SIG_IGN for SIGHUP")
```

/cc @jssjr @mleinart @drawks @brutasse

obfuscurity:
The cache and writer don't support log.debug, use log.msg instead. Fixes #401.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -71,7 +71,7 @@ class SortedStrategy(DrainStrategy):
       while True:
         t = time.time()
         metric_counts = sorted(self.cache.counts, key=lambda x: x[1])
-        log.debug("Sorted %d cache queues in %.6f seconds" % (len(metric_counts), time.time() - t))
+        log.msg("Sorted %d cache queues in %.6f seconds" % (len(metric_counts), time.time() - t))
         while metric_counts:
           yield itemgetter(0)(metric_counts.pop())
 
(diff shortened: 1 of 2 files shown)


obfuscurity:
@jssjr I think it would make sense long-term to have a working `log.debug()` in master, but I don't think any of these changes mandate its use. Merging this in.

PR #404, opened 2015-04-21
Clear buffers when they are reset.

liamstewart:
The aggregator can use multiple buckets in an aggregation computation. It marks
each bucket as inactive after the computation, but buckets can be reactivated
if new points are added - this can occur if the aggregation computation timing
doesn't align with the bucket boundaries. If that happens, then points will be
re-used between computations, leading to inflated or otherwise incorrect
values. Cleaning buckets after they are used ensures that old data is never
used again even though the bucket may be reused.

### DIFF
--- a/lib/carbon/aggregator/buffers.py
+++ b/lib/carbon/aggregator/buffers.py
@@ -96,6 +96,7 @@ class IntervalBuffer:
     self.active = True
 
   def mark_inactive(self):
+    self.values = []
     self.active = False
 
 


obfuscurity:
Maybe I'm confused, but if we're recalculating aggregation buckets, wouldn't we want the original values reused with the new datapoints? Once it's flushed it should overwrite the existing timestamp/value in Whisper.

mleinart:
This looks to break the use case where WRITEBACK_FREQUENCY is less than the frequency of aggregation (e.g. you want to flush your hourly aggregate every minute so that you always have some data for your latest point).

I think the same thing can be achieved by setting MAX_AGGREGATION_INTERVALS to 1 (note: this means if a datapoint comes in older than (now - aggregation_frequency) it'll be ignored in the aggregate)

stravadeploy:
Yup, I missed the line where the timestamp of the bucket is re-used. Please close this PR.

liamstewart:
Yup, I missed the line where the bucket timestamp is re-used. Thanks for the double-check.

Issue #324, opened 2014-11-06
Twisted backlog


juise:
Hi!

I have experienced some problems related to the hard coded tcp socket backlog size which opened by Twisted. I can't tune it from setting, and I have to fix the default backlog size in the code.

obfuscurity:
Fixed by #405.

PR #405, opened 2015-04-23
Tunable tcp backlog

jssjr:
Implements a config setting that exposes the TCP backlog parameter used when creating the listen socket for plain-text and pickle TCPServers. Also implements the backlog for the cache query server.

Note that this increases the default backlog to 1024, which in my opinion is a more realistic value for modern server hardware.

Fixes https://github.com/graphite-project/carbon/issues/324


jssjr:
Before:

```
metrics-carbon1-cp1-stg.iad /proc/12556 # ss -lt  | grep 2113
LISTEN     0      50                      *:2113                     *:*
```

After:

```
metrics-carbon1-cp1-stg.iad /proc/12556 # ss -lt  | grep 2113
LISTEN     0      1024                    *:2113                     *:*
```

mleinart:
:thumbsup:

yuuki1:
:+1:

syepes:
+1 This os a must have

PR #407, opened 2015-04-23
Skip caching of aggregation transformation output

mleinart:
This was originally merged into master and forgotten about. This pull cherry-picks the change into 0.9.x

Reported in issue #185 via @dkulikovsky
Each aggregation rule caches transformations to avoid unneeded regex
matches. In the case of many rules, each rule will maintain a cache for
every metric it sees, whether or not that metric is matched. This
manifests for some users as a memory leak

Instead, compromise by not caching the negative result, storing only
metric transformations that actually match have an output

### DIFF
--- a/lib/carbon/aggregator/rules.py
+++ b/lib/carbon/aggregator/rules.py
@@ -93,7 +93,8 @@ class AggregationRule:
       except TypeError:
         log.err("Failed to interpolate template %s with fields %s" % (self.output_template, extracted_fields))
 
-    self.cache[metric_path] = result
+    if result:
+      self.cache[metric_path] = result
     return result
 
   def build_regex(self):


obfuscurity:
Thanks!

jraby:
This fix manifests itself has a very noticeable performance drop for aggregators which see lots of metrics and only a small(ish) fraction of those match the rules.

I ran the following simple benchmark to see what the perf hit was:
 - carbon-aggregator configured with 16 rules.
 - send a fixed number of datapoints matching 8 of those rules (~17M datapoints)
 - measure how long it takes to pass through the aggregator

image

I was about to suggest an LRU cache and code it up, but I see that it was already done in https://github.com/graphite-project/carbon/pull/335.

I'll test that patch and post the benchmark result.

jraby:
Well, the LRU cache didn't perform up to my expectations :(

Running the same test as above:
image

That's half the performance of the "no negative cache" fix.

According to kernprof, adding a key and getting the value of a key (while maintaining the key order) seems to be very costly:
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    20                                               @profile
    21                                               def __contains__(self, key):
    22   1922752      1122547      0.6    100.0          return key in self.cache

File: /opt/graphite/lib/carbon/aggregator/rules.py
Function: __getitem__ at line 24
Total time: 12.9574 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    24                                               @profile
    25                                               def __getitem__(self, key):
    26   1921856      6895675      3.6     53.2          value = self.cache.pop(key)
    27   1921856      5284876      2.7     40.8          self.cache[key] = value
    28   1921856       776865      0.4      6.0          return value

File: /opt/graphite/lib/carbon/aggregator/rules.py
Function: __setitem__ at line 30
Total time: 0.007041 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    30                                               @profile
    31                                               def __setitem__(self, key, value):
    32       896          425      0.5      6.0          try:
    33       896         2517      2.8     35.7              self.cache.pop(key)
    34       896          617      0.7      8.8          except KeyError:
    35       896          593      0.7      8.4              if len(self.cache) >= self.capacity:
    36                                                           self.cache.popitem(last=False)
    37       896         2889      3.2     41.0          self.cache[key] = value

File: /opt/graphite/lib/carbon/aggregator/rules.py
Function: get_aggregate_metric at line 114
Total time: 28.2518 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   114                                             @profile
   115                                             def get_aggregate_metric(self, metric_path):
   116   1922752      7340618      3.8     26.0      if metric_path in self.cache:
   117                                               #if self.cache.has_key(metric_path):
   118   1921856     20894746     10.9     74.0        return self.cache[metric_path]
   119
   120       896         1590      1.8      0.0      match = self.regex.match(metric_path)
   121       896          501      0.6      0.0      result = None
   122
   123       896          464      0.5      0.0      if match:
   124        49          141      2.9      0.0        extracted_fields = match.groupdict()
   125        49           25      0.5      0.0        try:
   126        49          124      2.5      0.0          result = self.output_template % extracted_fields
   127                                                 except TypeError:
   128                                                   log.err("Failed to interpolate template %s with fields %s" % (self.output_template, extracted_fields))
   129
   130       896        13122     14.6      0.0      self.cache[metric_path] = result
   131       896          499      0.6      0.0      return result
```


So I went looking around for a faster implementation and found https://github.com/amitdev/lru-dict.

With this, I get better results (and reduced memory usage compared to the python LRU):
image



Both side by side:
image

So, to sum up:
- ~50k metrics/s with the original code (cache everything, forever)
- ~43k metrics/s with the C LRU
- ~32k metrics/s with the current code
- ~15k metrics/s with the LRU based on OrderedDict from #335 

For our use case, we would very much like to get those 11k metrics per second back :-)
(In fact we'd rather get the whole 50k metrics per second... but that's for later)

obfuscurity:
I would be happy to see any patches that improve performance, but we're not going to revert a change which fixes memory leaks for some users. I'm not sure that's what you're proposing, but I just wanted to rule it out. :wink:

jraby:
I'm not proposing anything yet since I don't have a good solution, but I wanted to highlight the cost of the current code and the (non) performance of the fix proposed in #335 .

I spent a fair bit of time yesterday trying to make this faster, but couldn't get past 50k metrics/s.
On my test rig, carbon-cache saturates at around 130k metrics/s, I find it depressing that the aggregator saturates well below that.

offtopic: @obfuscurity have you ever tried running those 2 programs under pypy?  I did that yesterday before giving up. The results were interesting... 
- carbon maxed out at around 600k metrics/s.
- the aggregator with the current code stayed at ~ 35k lines/s
- using the original code  yielded ~115k metrics/s. 
- using the 'slow' LRU code from #335 yielded ~100k metrics/s.
- I couldn't test with lru-dict, it segfaulted and I didn't have symbols to debug...

I'll spend some time testing that properly, but that avenue looks much more promising than trying to micro optimise the code, for our use case at least.

Issue #371, opened 2015-01-27
carbon.agents.cache.size always increasing in 0.9.13pre1


spencert:
I am running the 0.9.13pre1 (specifically 36749eedce6be908109d5133bc173721d9662733) and I am noticing that the carbon.agents.*.cache.size is always increasing. 

My setup is as follows. 2 physical servers, both with 32GB RAM and 10x15K SAS disks in a RAID10. For the metric load I am sending the servers disks keep up just fine.

All metrics come into server 1 which replicates metrics to itself (0.9.12 carbon) and server 2 (0.9.13pre1 carbon) over plain-text using carbon-c-relay. The carbon.conf for the [cache] section are exactly identical and I am running two caches on each physical server. As you can see I limit the MAX_UPDATES_PER_SECOND = 500 for each cache which makes sense on this hardware. However it is clear that 0.9.13pre1 cache.size always increases whereas 0.9.12 does not.

I have attached two images which are the exact same metric targets just different hosts.

![0 9 12-carbon](https://cloud.githubusercontent.com/assets/947938/5920539/9276de3c-a5fa-11e4-8ebb-43df8fbf3b18.png)
![0 9 13pre1-carbon](https://cloud.githubusercontent.com/assets/947938/5920702/c47feb5c-a5fb-11e4-90bf-3ec0a444ac90.png)

obfuscurity:
Not that I dispute your findings, but the cache.size on the 2nd graph certainly looks like a counter and not a derivative. Can you double-check your target definition there?

spencert:
Sure thing. I thought the same thing, here is the actual URL (hostname scrubbed).

**0.9.13pre1**
http://db5234.$dayjob.com/render/?width=586&height=308&_salt=1422370542.946&from=-14days&target=alias%28sumSeries%28carbon.agents.*.cache.queues%29%2C%22cache.queues%22%29&target=alias%28secondYAxis%28sumSeries%28carbon.agents.*.cache.size%29%29%2C%22cache.size%22%29&target=alias%28sumSeries%28carbon.agents.*.metricsReceived%29%2C%22Metrics%20Received%22%29&target=alias%28sumSeries%28carbon.agents.*.updateOperations%29%2C%22Update%20Operations%22%29&title=0.9.13pre1%20Carbon

**0.9.12**
[...]

PR #408, opened 2015-04-24
Fix MetricsCache size leak in 0.9.x

jssjr:
Exploration into a suspected leak in MetricsCache.

Initial indication of a problem came from inspecting the running carbon-cache process.

```
>>> print "%d %d %d" % ( len(MetricCache), MetricCache.size, reduce(lambda x, y: x + len(y), MetricCache.values(), 0) )
4771 5157 5355
```

I started by backporting tests from `master` then working from there, eventually adding some benchmarks available through `tox -e benchmark`. By reorganizing how we handle instrumentation and event includes, I was able to resolve the problem I discovered.

screenshot 4 24 15 9 57 pm

I'm still verifying this behavior on our staging cluster, but things are looking quite good.

/cc https://github.com/graphite-project/carbon/issues/371 because this fixes the behavior seen there.

### DIFF
--- a/lib/carbon/instrumentation.py
+++ b/lib/carbon/instrumentation.py
@@ -116,9 +116,9 @@ def recordMetrics():
   # aggregator metrics
   elif settings.program == 'carbon-aggregator':
     record = aggregator_record
-    record('allocatedBuffers', len(BufferManager))
+    record('allocatedBuffers', len(carbon.aggregator.buffers.BufferManager))
     record('bufferedDatapoints',
-           sum([b.size for b in BufferManager.buffers.values()]))
+           sum([b.size for b in carbon.aggregator.buffers.BufferManager.buffers.values()]))
     record('aggregateDatapointsSent', myStats.get('aggregateDatapointsSent', 0))
 
   # relay metrics
@@ -185,4 +185,3 @@ class InstrumentationService(Service):
 
 # Avoid import circularities
 from carbon import state, events, cache
-from carbon.aggregator.buffers import BufferManager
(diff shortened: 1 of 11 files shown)


obfuscurity:
so_beautiful

obfuscurity:
dancing

jssjr:
I am still verifying that all of the instrumentation is working as designed. Please do not merge this yet.

lib/carbon/instrumentation.py
```diff
@@ -116,9 +116,9 @@ def recordMetrics():
   # aggregator metrics
   elif settings.program == 'carbon-aggregator':
     record = aggregator_record
-    record('allocatedBuffers', len(BufferManager))
+    record('allocatedBuffers', len(carbon.aggregator.buffers.BufferManager))
```
jssjr: I suspect this won't work. Instead, I'll likely have to rework how this stat is instrumented.

deniszh:
I tried [that solution](https://github.com/graphite-project/carbon/issues/371#issuecomment-118056194), works fine, worth to check.

obfuscurity:
@jssjr Are we ok to merge this?

obfuscurity:
@jssjr Can we get the patch from https://github.com/graphite-project/carbon/issues/371#issuecomment-118056194 added to this PR and finally get this merged?

deniszh:
@obfuscurity: Let me merge it in new PR

obfuscurity:
@deniszh ok, I'm going to merge this one. I assume you mean that you'll submit the #371 patch in a separate PR.

deniszh:
Yep, that'll be much easier.

obfuscurity:
Thanks for all your hard work on this one, @jssjr. Happy to see it in the tree.

PR #409, opened 2015-04-24
New tests for carbon aggregator buffers

mleinart:
Just a bunch of new tests for carbon.aggregation.buffers

Also a very minor regression uncovered by new tests: if WRITE_BACK_FREQUENCY is set, it should only be used if it's less than the aggregation frequency. If we ever want to support writing *less often* than the lowest aggregation frequency (i.e. periodic flushes of multiple aggregates) more code changes are needed.


SEJeff:
Yay for more tests

PR #419, opened 2015-07-12
tuned cache strategy & cache persistance

llicour:
Hello

Here are a set of patchs that add 2 majors features to carbon cache daemon, that improve performance and resiliance :
* write cache memory to disk at shutdown and regular interval to minimize unflushed metrics lost if daemon is killed. The file is automatically loaded at startup
* new tuned cache  write strategy that allow to focus flushing queues based on the size, name, age... Flushing occurs for a time or a number or percent of the queue length

This allow me to manage many queues and metrics without having a great storage backend (no more need of SSD storage)
This strategy consume more memory, but allow the system to be more responsive (less CPU & I/O)

See Git comments for some details

Also : A PR has been pushed on graphite-api GIT to add new APIs to dialog with this patch : 
(https://github.com/graphite-project/graphite-web/pull/1277)
* force to write cache memory to disk
* force queue flushing of some queues matching a regexp


deniszh:
@llicour
Quite interesting patch,  thanks! Could you please explain how new cache strategy helps you with slower storage?

llicour:
Ok, here is the big picture

When carbon cache receive more metrics it can process (that is the case if there is a slow storage backend), metrics are queued in memory (in a dictionnary structure) and dequeued by a thread, using a specific strategy

At this point, in memory, we have array of metrics, associated to a dictionnary key (the metric name)
Without dequeuing thread, system memory will be consumed until Out of memory (via swapping..)

Goal is to dequeing as fast as possible, considering the backend storage.
Dequeuing is done queue per queue, (a queue equivalent to a whisper db file .wsp)
There is a cost to process a whisper file. Using update_many() api allow to optimize I/O by updating DB file with many metrics in one operation
Consider that updating a whisper DB with 100 metrics can be equivalent than updating it with 1 metric, as 1 metric is only a few bytes

We can consider that for a specific hardware, the rate of queue/second is something that is fixed (it's depend of the hardware, and can be evaluated using the carbon metrics)
It's possible to limit this rate using the MAX_UPDATES_PER_SECOND, and decreasing this value allow to have a responsive system, as it consume less I/O and CPU
However, memory will growing, as we decrease the dequeing capability

As the rate queue/sec is fixed, if we want to limit memory consumption, we have to work on the rate metrics/sec  
Especially with a slow storage (but not only), it's better to have many metrics per queue (whisper DB) to update (rather than few) as it optimize this rate

Native strategies are not (well) design for this
* naive : it works randomly over the queue. Nothing to do with it
* sorted : the list of queues in memory are sorted by the number of metrics. Excellent ! So the thread process first queues with lots of metrics (that is a good idea). Pb is that all the queues are processed (even with few metrics). So if you have many queues, it will take time (hours...) before to process all the queues.
* max : this only select THE queue that contains the more metrics. Nice try. But if there is a lot of queues  in memory, the CPU is only busy to sort the queues (that takes seconds), for 1 whisper DB update. Unusable...

Idea of a new strategy was born here : Having a way to only select queues that worth to be flushed, by processing a 'sorted'  strategy only for the top 100 (for example)
Tuned strategy also allow to have more criteria to select queues

One of the drawback of this strategy is that queues can be keeped in memory for long time before beeing flushed (on my system, this can be days). CarbonLink queries can take care of this, so this is not a design problem
However, there is a problem if carbon-cache process is shutdown or system reboot, as all queues in memory will be lost. Also,during a clean shutdown, it can take a long time (hours) before all the queues to be flushed.
For these reasons, the patch also add a mecanism to flush the cache memory in a flat file (grep-able and awk-able) at shutdown, that is reloaded during startup. It takes only few seconds to create this file...

I also think that even with a speedy storage (SSD), it can be valueable to use this strategy (decreasing the MAX_UPDATES) to limit the micro-I/O that can be done on the SSD drive (that could tired the disk)

Hope these explanations (in my own forked english) could help you ...

Laurent

deniszh:
Thanks,  Laurent , excellent explanation!  Found your commit comments too. Will try your patch soon on our cluster.

llicour:
ok.
So wait for few days, as i'll push commit that add metrics about tuned strategy, as well as graphite api that allow to modify strategie params online (without restarting cabon-cache)

llicour:
I've just update the PR to include 
* new carbon.agents metrics that allow to follow 'tuned' strategy processing
* new CarbonLink command that allow to change some carbon cache settings online

The following settings are supported :
 - CACHE_WRITE_STRATEGY
 - CACHE_WRITE_TUNED_STRATEGY_LARGEST
 - CACHE_WRITE_TUNED_STRATEGY_RANDOM
 - CACHE_WRITE_TUNED_STRATEGY_OLDEST
 - CACHE_PERSIST_INTERVAL
 - MAX_UPDATES_PER_SECOND
Sample :
curl localhost/metrics/set-param -X POST -d '{"key":"CACHE_WRITE_STRATEGY","value":"tuned"}'
curl localhost/metrics/set-param -X POST -d '{"key":"MAX_UPDATES_PER_SECOND","value":"1"}'

Note: changes are not persistant. If process is restarted, settings from configuration file will be restored

frlinux:
Hey guys, did this make it to the 0.9.13 tarballs released last week?

deniszh:
0.9.13 still not released, as far as I know. And this PR is not even in
milestone, so...
I tested cache persistence on my own branch (maybe I need to open source
it) but didn't test new strategy...

frlinux:
Thanks for the clarifications. Looks like the tarballs for webui and carbon are now gone from http://graphite.readthedocs.org/en/latest/releases/0_9_13.html

obfuscurity:
I think this is a very interesting patch but it really belongs in master, not the 0.9.x branch. @llicour would you be so kind as to rework this patch against master Carbon and Graphite-Web?

llicour:
No problem for working on this, but i have to install a plateform from the master branch.

Could you clarify the roadmap about graphite/carbon branch ?
It seems to me that the 0.9.x branch is the current stable version (latest version published on pypi)
and that master branch is an experimental branch (a complete rewrite) that does not have been published yet

What is the roadmap of each of these branches (main features, next release dates...) ?
Is the 0.9.x branch a dead branch ?

obfuscurity:
The 0.9.x branch was forked a long time ago, I don't even remember why at this point (predates my involvement). We've nursed it along as the "stable" branch, mostly applying security patches & bugfixes and cherry-picked features from master deemed too important to skip. All new feature development should always be targeted at the master branch.

The Graphite Project is entirely run by volunteers. As such it's impossible (or at best, disingenuous) to give release dates in advance. My off-the-cuff answer to "when is the next release cut from master" would be something along the lines of "probably 3-6 months after our next (and final) release from 0.9.x", which should be any day or week now.

obfuscurity:
@llicour As per my comment in https://github.com/graphite-project/graphite-web/pull/1277, this looks really cool but please resubmit against master. Thanks! :heart:

cchiremath:
Hi ...
When this will be merged to master. i checked in 9.15 but i dont see the changes .

please help with the info.

deniszh:
It was never merged to 0.9.x and an author does not submit it to master.
I personally using part of this patch in PROD and will port it to master soon.
Stay tuned. :)

deniszh:
Just a status update:
1) patch is quite harsh with twisted reactor. If flush or persisted cache save take more than 15-20 seconds reactor going a bit mad, and causing carbon cache reconnect with relay. Fixed it with increasing cache on relay on my working 0.9.x installation, but it's still quite scare.
2) porting this patch to master is also not straightforward. Tried that - bit no luck yet. Will try to create my own solution, somewhat later.

cchiremath:
i tried to make the build on centOS 6, i get following error.
Compiling /builddir/build/BUILDROOT/graphite-web-0.10.0_rc1-1.x86_64/opt/graphite/webapp/graphite/remote_storage.py ...
SyntaxError: ('invalid syntax', ('/opt/graphite/webapp/graphite/remote_storage.py', 177, 13, ' for series in unpickle.loads(pickled_response)\n'))

plz help.

deniszh:
Sorry, @cchiremath, not really sure what you trying to compile. This patch is against 0.9.x branch, you can't apply it on a master branch - there are too many changes.

PR #425, opened 2015-09-01
Fix the logrotate config variable name

benburry:
The setting variable that controls whether carbon is doing its own logrotation is called ENABLE_LOGROTATE.

This was incorrectly included in the example carbon config as ENABLE_LOGROTATION.

Also fix failing tests (caused by updated version of mock)


obfuscurity:
Actually we can see that it was intentionally changed to `ENABLE_LOGROTATION` in https://github.com/graphite-project/carbon/commit/4793eb90bc58f55b066536a5fdf15114ba974fea for 0.9.x. This should be changed to match in master.

PR #427, opened 2015-09-04
Name internal logrotation setting variable consistently

benburry:
The internal logrotation setting was renamed for the 0.9.x version from ENABLE_LOGROTATE to ENABLE_LOGROTATION, but the changes were only partially applied to master.


obfuscurity:
Thanks!

PR #429, opened 2015-09-11
Timeout idle client connection

pu239ppy:
Work originally based on @piontas Close connection after n seconds if no data is received #367


obfuscurity:
I'm going to go ahead and merge it, then make a small change to include the example for all daemon types.

PR #437, opened 2015-09-12

obfuscurity:
Introduced by @pratX in #231. I love the feature, but it needs some more eyes on it. My primary concern is the possible confusion between `LOW_CACHE_SIZE` and `CACHE_SIZE_LOW_WATERMARK`.


obfuscurity:
@pratX the goal of this patch is to increase batch writes, yes? In that case, is it materially any different than lowering `MAX_UPDATES_PER_SEC`?

/cc @drawks @mleinart @jssjr

pratX:
@obfuscurity With max or sorted write strategy lowering the MAX_UPDATES_PER_SEC works well. 
With unordered write strategy, coupling lower MAX_UPDATES_PER_SEC  with MIN_DATAPOINTS_PER_UPDATE gives better throughput.

obfuscurity:
@pratX What do you think of the comment changes I've just added? I think this should help educate users when to use this feature.

pratX:
@obfuscurity  there are bugs in the current version; I have forked obfuscurity/carbon, would be fixing the bugs in feature/min_datapoints_per_update branch, test it and then give you a pull request.

obfuscurity:
I don't understand the point of all the added complexity in 5f846f743b2776868f9e9f3cd482c6f8a174dd19. Why are we suddenly doing all this file access directly instead of via CarbonLink?

pratX:
At Media.Net we have many per day metrics, and we plot their aggregate. When I set MIN_DATAPOINTS_PER_UPDATE to 10, the new metrics took 10 days to reflect in the graphs. So, I added the logic to waive min_number_of_datapoints constraint for new metrics. But as @obfuscurity  has mentioned, a better way to do this is to make graphite-web look for a metric in carbon-cache even if there is no whisper file for it. So, I'll be removing the logic for special handling of new metrics from DrainStrategy.

GoozeyX:
Is there an update to this on whether or not this is going to be implemented?

obfuscurity:
FML

obfuscurity:
Sorry folks, moving this one over to #580.

PR #439, opened 2015-09-12
Expire old aggregated datapoints

obfuscurity:
Introduced by @plankthom in #176, this adds `MAX_AGGREGATION_AGE` for controlling how long to store aged datapoints in aggregator memory before expiring them (e.g. no new datapoints are arriving).

Also, feels like an alternative / less intrusive approach to the problem addressed in https://github.com/graphite-project/carbon/pull/335.

This PR specifically cleans up a typo and some whitespace bits from the original PR.


obfuscurity:
This breaks a lot of our tests because `most_recent_interval` is still initialized to `None` when called in the mathematical operations in L55 and L68. /cc @plankthom

PR #445, opened 2015-09-20
Support for direct AMQP exchange

obfuscurity:
Introduced by @milosgajdos83 in #338, this PR fixes the merge conflict. From the original description:

```
With this PR you have an option to run multiple carbon-relays which can
essentially behave as "workers". These carbon "workers" are reading in
metrics from the direct AMQP exchange and distribute them to carbon-caches
or to other relays based on relay method - there is no binding key necessary.
Previously you could only distribute based on binding/routing key
so you were not able to have true relay WORKERS because having them
running like that would generate duplicate writes on your drive.
With topic exchange and default binding key '#' you'd always get duplicate writes.
Direct exchange allows metric to be consumed by ONE carbon cache worker - the
mirroring is then handled vi carbon itself by specifying a particular carbon configuration.
```


obfuscurity:
@milosgajdos83 Please test this one.

milosgajdos83:
Excellent. Will have a go at this at some point this week. Thanks.

milosgajdos:
Wow! Blast from the past!

PR #446, opened 2015-09-24
Add unit tests for storage module

JeanFred:
As I was trying to track down either a bug in Carbon parsing or a mistake in my `storage-aggregation.conf` file (for the story, it turned out my carbon deamon just had to be restarted :-°), I ended up writing this test suite for the storage module.

The all `NoConfigSchemaLoadingTest` class, whose purpose is to test what happens when `CONF_DIR` leads nowhere, is commented out though. For some reason, the mock of `carbon.conf.settings` persists between classes. (if I test  `NoConfigSchemaLoadingTest` by itself it works fine). I have fiddled around with trying to `reload(carbon.conf)` in setUp or tearDown but to no avail. Maybe someone will have a better idea? Otherwise I suppose it’s fine removing this test class alltogether.

Hope that helps!

### DIFF
--- a/lib/carbon/tests/test_storage.py
+++ b/lib/carbon/tests/test_storage.py
@@ -1,0 +1,107 @@
+import os
+from unittest import TestCase
+from mock import patch
+
+
+# class NoConfigSchemaLoadingTest(TestCase):
+
+#     def setUp(self):
+#         settings = {
+#             'CONF_DIR': '',
+#         }
+#         self._settings_patch = patch.dict('carbon.conf.settings', settings)
+#         self._settings_patch.start()
+
+#     def tearDown(self):
+#         self._settings_patch.stop()
+
+#     def test_loadAggregationSchemas_load_default_schema(self):
+#         from carbon.storage import loadAggregationSchemas, defaultAggregation
+#         schema_list = loadAggregationSchemas()
+#         self.assertEquals(len(schema_list), 1)
+#         schema = schema_list[0]
+#         self.assertEquals(schema, defaultAggregation)
+
+#     def test_loadStorageSchemas_raise_CarbonConfigException(self):
+#         from carbon.storage import loadStorageSchemas
+#         from carbon.exceptions import CarbonConfigException
+#         with self.assertRaises(CarbonConfigException):
+#             loadStorageSchemas()
+
+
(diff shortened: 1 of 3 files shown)


JeanFred:
Thoughts? :)

lib/carbon/tests/test_storage.py
```diff
@@ -0,0 +1,107 @@
...
+#         schema_list = loadAggregationSchemas()
+#         self.assertEquals(len(schema_list), 1)
+#         schema = schema_list[0]
+#         self.assertEquals(schema, defaultAggregation)
+
+#     def test_loadStorageSchemas_raise_CarbonConfigException(self):
+#         from carbon.storage import loadStorageSchemas
+#         from carbon.exceptions import CarbonConfigException
+#         with self.assertRaises(CarbonConfigException):
+#             loadStorageSchemas()
```
SEJeff: Please don't add large swaths of code that is commented out. Uncomment it or remove it.

SEJeff:
More tests are always great, although I'm not the biggest fan of having to do the imports in every test. DRY

Please remove or fix the commented code and we will happen my merge this however. Thanks for doing this!

lib/carbon/tests/test_storage.py
JeanFred: As I explained in the PR, I was hoping someone might be able to enlighten me regarding this mock persistence across tests. If not, sure I’ll remove it.

SEJeff: Gotcha. I can look at it more tomorrow or this weekend

JeanFred:
> More tests are always great, although I'm not the biggest fan of having to do the imports in every test. DRY

I *think* I had to do this because of the mock persistence, although I am not entirely sure anymore.

SEJeff:
@JeanFred Thanks that makes sense. Merged! Feel free to write more if you feel so inclined. Sorry this took so long to get merged

JeanFred:
Thanks for merging @SEJeff! What about the commented out code − shall we just leave it there until someone figures it out ? Or let’s remove it as it will probably rot otherwise ?

SEJeff:
@JeanFred if I find the time I'll take a look at it, but carbon is a piece of the graphite project that I tend to not deal with much as I don't actually use it. @mleinart is really good wrt carbon

PR #448, opened 2015-09-27
Fixing MetricsCache size leak.

deniszh:
Created from @jjneely's comment https://github.com/graphite-project/carbon/issues/371#issuecomment-118056194


obfuscurity:
Thanks!

PR #452, opened 2015-10-04
Restore CH routing that replicas exist on different servers.

deniszh:
Making https://github.com/graphite-project/carbon/pull/135 optional, e.g. will not break routing for existing cluster but can be optionally enabled by user.


deniszh:
Please review - @dctrwatson @obfuscurity @SEJeff @jssjr 
Thanks!

conf/carbon.conf.example
```diff
@@ -262,6 +262,14 @@ RELAY_METHOD = rules
 # datapoint to more than one machine.
 REPLICATION_FACTOR = 1
 
+# If your REPLICATION_FACTOR >=2, you are using more than one replica per server
```
obfuscurity: Not trying to be an oxford comma groupie here, but I think we need one at the end of this sentence or it takes on a different meaning.

conf/carbon.conf.example
```diff
@@ -262,6 +262,14 @@ RELAY_METHOD = rules
...
+# and you want to be sure that every replica of data is located on different servers
```
obfuscurity: Same here, please add a comma at the end of the sentence.

conf/carbon.conf.example
```diff
@@ -262,6 +262,14 @@ RELAY_METHOD = rules
...
+# please setup DIVERSE_REPLICAS to True.
+# Before carbon v0.9.14 it was not guaranteed!
```
obfuscurity: I would prefer that we give an "active" explanation of what this feature actually does, rather than implying how it might or might not be different from past behavior.

Also, rather than saying `Before carbon v0.9.14`, let's describe what the past behavior would be in the context of `DIVERSE_REPLICAS` being `True` or `False`.

obfuscurity: e.g. "the equivalent behavior for Carbon 0.9.12 was `DIVERSE_REPLICAS = False` but before 0.9.11 it was `DIVERSE_REPLICAS = True`". Note that I have **no idea** what the actual behavior was, just giving examples.

obfuscurity:
Other than those documentation comments, LGTM at a glance. Have not tested.

deniszh:
Ok, let's finish that in comments first:
```shell
# If your REPLICATION_FACTOR >=2, and you are using more than one replica per server,
# and you want to be sure that every replica of data is located on different servers,
# please setup DIVERSE_REPLICAS to True.
# For example, if you have 
# DESTINATIONS = [ hostA:st1, hostA:st2, hostB:st1, hostB:st2 ] and
# REPLICATION_FACTOR = 2 and
# DIVERSE_REPLICAS = True
# then for any metric one replica will be present on hostA, another replica - on hostB.
# If DIVERSE_REPLICAS = False then it's possible that for some metrics both replicas will be 
# routed to hostA and you'll get no active replicas if host goes down.
# Beware - enabling it could breaks your _existing_ (pre v0.9.14) cluster routing!
# It's better to enable that for new clusters.
#
#DIVERSE_REPLICAS = False
```
Is that looks better?

deniszh:
Will edit comment for master too, if looks good.

obfuscurity:
I think we can simplify it a bit and still get the same message across. How does this sound?

```shell
# For REPLICATION_FACTOR >=2, set DIVERSE_REPLICAS to True to guarantee replicas
# across distributed hosts. With this setting disabled, it's possible that replicas
# may be sent to different caches on the same host. This has been the default 
# behavior since Carbon 0.9.???.
# Note that enabling this on an existing pre-0.9.14 cluster will require rebalancing
# your metrics across the cluster nodes using a tool like Carbonate.
#DIVERSE_REPLICAS = False
```

deniszh:
Ahem... Carbonate needs to be patched too, because by default it will use diverse_metricas=False - https://github.com/jssjr/carbonate/blob/1b49c332b274cfa208c00f5b94a29630eb11b91b/carbonate/cluster.py#L16
Will create PR there too. :)

obfuscurity:
Ouch.

obfuscurity:
cc @jssjr

deniszh:
I could make 'diverse_replicas=True' by default in code, and override it with config. Then Carbonate (and other tools which are using carbon's code) will work 'out-of-the-box' - but if someone upgrades cluster w/o reading documentation and introducing DIVERSE_REPLICAS = False in config - ouch...
So, maybe thats default is a good thing for master but for 0.9.x - not sure.

obfuscurity:
No, I think we definitely need to keep it disabled in 0.9.x, but we should look at enabling it in master since we're justified in introducing breaking changes there imho.

deniszh:
Ok, I'll 
1) fix comment to:
```shell
# For REPLICATION_FACTOR >=2, set DIVERSE_REPLICAS to True to guarantee replicas
# across distributed hosts. With this setting disabled, it's possible that replicas
# may be sent to different caches on the same host. This has been the default 
# behavior since introduction of 'consistent-hashing' relay method.
# Note that enabling this on an existing pre-0.9.14 cluster will require rebalancing
# your metrics across the cluster nodes using a tool like Carbonate.
#DIVERSE_REPLICAS = False
```
on 0.9.x and master.
2) enable diverse_replicas=True in code and in config for master
3) will create PR for DIVERSE_REPLICAS option for Carbonate

obfuscurity:
Thanks!

PR #453, opened 2015-10-04
Restore CH routing that replicas exist on different servers.

deniszh:
Backport of https://github.com/graphite-project/carbon/pull/452 to master.

Making https://github.com/graphite-project/carbon/pull/135 optional, e.g. will not break routing for existing cluster but can be optionally enabled by user.


obfuscurity:
I actually think we should enable this by default in master. If we're going to introduce a breaking change, this is the time/place to do it, right? :smile_cat:

conf/carbon.conf.example
```diff
@@ -248,6 +248,14 @@ RELAY_METHOD = rules
 # datapoint to more than one machine.
 REPLICATION_FACTOR = 1
 
+# For REPLICATION_FACTOR >=2, set DIVERSE_REPLICAS to True to guarantee replicas
+# across distributed hosts. With this setting disabled, it's possible that replicas
+# may be sent to different caches on the same host. This has been the default 
+# behavior since introduction of 'consistent-hashing' relay method.
+# Note that enabling this on an existing pre-0.9.14 cluster will require rebalancing
+# your metrics across the cluster nodes using a tool like Carbonate.
+DIVERSE_REPLICAS = True
```
obfuscurity: We can leave this commented out.

obfuscurity:
LGTM, thanks!

PR #458, opened 2015-10-11
Don't use mirrors for pip

alex:


obfuscurity:
Why? /cc @brutasse @SEJeff

alex:
The flag doesn't do anything on modern pips, and on older pips it doesn't actually help with uptime. /cc @dstufft

obfuscurity:
Ah ok, I see that now in their release notes. Shows you how carefully I pay attention to pip releases. :wink: Thanks!

brutasse:
:+1:

alex:
np, sorry I didn't explain the reason in the original PR.

Issue #82, opened 2013-03-11
whisper files created with no permissions


mleinart:
#### Imported from https://bugs.launchpad.net/bugs/702066
|||
|----|----|
|Reported by|[pete-appnexus](https://launchpad.net/~pete-appnexus)|
|Date Created|Jan 12, 2011|
I'm using carbon-relay (Graphite-0.9.6) to send metrics from one server to another.

Currently, I populate a metric on one server, and it looks fine in the browser.

I look on the other server, and I get this:

Traceback (most recent call last):
  File "/usr/lib/python2.4/site-packages/django/core/handlers/base.py", line 86, in get_response
    response = callback(request, *callback_args, **callback_kwargs)
  File "/opt/graphite/webapp/graphite/render/views.py", line 97, in renderView
    seriesList = evaluateTarget(target, timeInterval)
  File "/opt/graphite/webapp/graphite/render/evaluator.py", line 14, in evaluateTarget
    result = evaluateTokens(tokens, timeInterval)
  File "/opt/graphite/webapp/graphite/render/evaluator.py", line 30, in evaluateTokens
    return evaluateTokens(tokens.expression, timeInterval, originalTime)
  File "/opt/graphite/webapp/graphite/render/evaluator.py", line 43, in evaluateTokens
    dbResults = dbFile.fetch( timestamp(startTime), timestamp(endTime) )
  File "/opt/graphite/webapp/graphite/storage.py", line 232, in fetch
    (timeInfo,values) = whisper.fetch(self.fs_path, startTime, endTime)
  File "/usr/lib/python2.4/site-packages/whisper.py", line 462, in fetch
    fh = open(path,'rb')
IOError: [Errno 13] Permission denied: '/opt/graphite/storage/whisper/path/to/my/metric.wsp'

The permissions on the file are 0000 instead of 0644 as I'm seeing on the source server.

The directory used for this is a 20GB tmpfs, although I'm not sure that should play into it.

Have you seen anything like this before, and can you help me track this down?

Thanks,
Pete

JensRantil:
Any progress on this? Just hit this bug...

deniszh:
https://github.com/graphite-project/carbon/issues/82#issuecomment-52194451

jacobrichard:
[...]

PR #459, opened 2015-10-12
Monkeypatch twisted logger to fix thread/umask bug

obfuscurity:
This is a proposed fix from @jacobrichard for the [Twisted umask race bug](https://twistedmatrix.com/trac/ticket/7026) as reported in #82. It would be **really** nice to finally squash this 5+ year-old bug, once and for all.

**PLEASE TEST THIS CHANGE AND REPORT BACK**

### DIFF
--- a/lib/carbon/log.py
+++ b/lib/carbon/log.py
@@ -15,6 +15,26 @@ class CarbonLogFile(DailyLogFile):
     from carbon.conf import settings
     self.enableRotation = settings.ENABLE_LOGROTATION
 
+  def _openFile(self):
+    """
+    Fix Umask Issue https://twistedmatrix.com/trac/ticket/7026
+    """
+    openMode = self.defaultMode or 0777
+    self._file = os.fdopen(os.open(
+      self.path, os.O_CREAT|os.O_RDWR, openMode), 'r+', 1)
+    self.closed = False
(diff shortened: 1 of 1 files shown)


ShalomCohen:
I couldn't make it work on 0.9.10, it seems exactly the same as the 0.9.13 file - are you sure this one is backward compatible - I just mentioned it on the 0.9.13 commit thread as well.

jacobrichard:
In a test pulling from the current master branch, I see some issues with missing imports for the 'os' library. That was a miss on my part in my commit, sorry for the confusion.

The relevant code is in ba2328d097ebf41672711f429f0348b603135adb

obfuscurity:
@jacobrichard Sorry, I don't see the problem with the way we're importing `exists`. Can you elaborate?

```python
Python 2.7.10 (default, Jul  9 2015, 13:34:07)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from os.path import exists
>>> exists('LICENSE')
True
>>> exists('foo')
False
```

obfuscurity:
@jacobrichard Disregard, I see what you mean. Fixing.

PR #463, opened 2015-11-05
more explicitly stating what 'creates' is, and that excess metrics ma…

gwaldo:
…y be discarded


obfuscurity:
We have one patch to master in #462 and this one here against 0.9.x, but they're not the same.

PR #476, opened 2015-11-25
Fix STORAGE_DIR

### LINKED ISSUE #269: STORAGE_DIR in carbon.conf has no affect
https://github.com/tmm1/graphite/blob/master/carbon/conf/carbon.conf.example#L14

Uncommenting L17 and changing the path has no affect. The only way I could find to change this path was by setting it as an environment variable. Not sure what's more wrong here, the example config, or if there is a bug in graphite/conf.py

reynir:
Some settings' defaults refer to STORAGE_DIR, e.g PID_DIR. These are
setdefault'ed using the STORAGE_DIR from before the config is read, i.e.
the defaults. The settings that depend on STORAGE_DIR must be
setdefault'ed each time the config file has been read.

This should fix #269 

It is perhaps not the best style. My initial fix was to move the STORAGE_DIR dependent setdefault's to the bottom of the function, but I found out some code above used LOG_DIR and PID_DIR.

I don't feel this is the correct solution, but it seems to fix #269 for now. At least it shows the cause(!)

### DIFF
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -547,15 +547,19 @@ def read_config(program, options, **kwargs):
                         os.environ.get("GRAPHITE_STORAGE_DIR",
                                        join(graphite_root, "storage")))
 
-    # By default, everything is written to subdirectories of the storage dir.
-    settings.setdefault(
-        "PID_DIR", settings["STORAGE_DIR"])
-    settings.setdefault(
-        "LOG_DIR", join(settings["STORAGE_DIR"], "log", program))
-    settings.setdefault(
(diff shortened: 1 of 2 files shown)


reynir:
Another solution could be to override __getitem__ in Settings

reynir:
The test cases in 2a3fff763955caf70401c15ded88f3f81d99be12 may be worth cherry-picking.

obfuscurity:
I might take another stab at refactoring this later but it's worth it to get this in now (and for those tests, yay!). Thanks for this contribution. :cake:

PR #480, opened 2015-11-30
Add LOG_CACHE_QUEUE_SORTS to master

obfuscurity:
This setting was never ported to master, let's address that.


obfuscurity:
The builds won't show completely clean since Python 2.6 has been deprecated. I've removed those from future builds but Travis won't recognize this for an existing build. Merging.

PR #483, opened 2015-12-02
changed the "Sorted" log messages into debug messages

LionelCons:
Our "console.log" gets full very quickly with messages like:

02/12/2015 08:50:30 :: Sorted 349 cache queues in 0.000427 seconds

These should be treated as debug messages.


obfuscurity:
We're addressing this in https://github.com/graphite-project/carbon/pull/480.

PR #484, opened 2015-12-03
Cherry-pick TimeSeriesDatabase from megacarbon branch

iain-buclaw-sociomantic:
It is becoming quite a pet peeve of mine to be stuck on the abandoned ``megacarbon`` branch and having to manually merge in upstream fixes.  :-)

While the ``megacarbon`` branch does a lot more, it is really only this part that is really relevant (to me), where all database actions are abstracted away, allowing any other database implementation to be plugged in.

I've stopped short of duplicating ``parseRetentionDefs`` from whisper and adding a ``ceres`` backend.  If there is no object against such things, I will push forward with those too.  However what's been changed here is enough to allow a pause for review.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -30,6 +30,10 @@
 #
 #LOCAL_DATA_DIR = /opt/graphite/storage/whisper/
 
+# Specify the database library used to store metric data on disk. Currently
+# only the whisper database is supported, but ceres support is coming soon.
+DATABASE = whisper
+
 # Enable daily log rotation. If disabled, a new file will be opened whenever the log file path no
 # longer exists (i.e. it is removed or renamed)
 ENABLE_LOGROTATION = True
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -21,8 +21,8 @@ from os.path import join, dirname, normpath, exists, isdir
 from optparse import OptionParser
 from ConfigParser import ConfigParser
 
-import whisper
-from carbon import log
+from carbon import log, state
+from carbon.database import TimeSeriesDatabase
 from carbon.exceptions import CarbonConfigException
 
 from twisted.python import usage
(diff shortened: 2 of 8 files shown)


mleinart:
Thanks, I'll try and take a look at this today

iain-buclaw-sociomantic:
OK, thanks.

FYI, I've finished the conversion here: https://github.com/graphite-project/carbon/compare/master...iain-buclaw-sociomantic:tsdatabase2

Doing some small tests and it all works fine with both whisper and ceres.

iain-buclaw-sociomantic:
Rebased, and fiddled with the ``getFilesystemPath`` test until it passed.

mleinart:
Note: when I say "I'll look at this today" it often means "I'll start feeling guilty about not looking at this starting today" ;)

I promise I haven't forgotten about this. This is something that has needed to be done for some time, thanks for the work!

mleinart:
I can kindof go either way with the change to TimeSeriesDatabase.create()'s method signature from (self, metric, **options) to specifying the options explicitly, but I'm just going to ship as is.

FYI, I'd like to hold off merging Ceres from the other branch until graphite-web supports reading past retentions which is something I've been working on getting put in

mleinart:
Merged in eeba611

iain-buclaw-sociomantic:
> Note: when I say "I'll look at this today" it often means "I'll start feeling guilty about not looking at this starting today" ;)

I am on holidays so there is no rush on my side, yet. :-)

> I can kindof go either way with the change to TimeSeriesDatabase.create()'s method signature from (self, metric, **options) to specifying the options explicitly, but I'm just going to ship as is.

In megacarbon, there is a ``determine_metadata`` function which returns a value to pass as this param.  I'm not really fond of the signature, but when considering any of the following.
- The absence of any function/method to return 'determined' metadata that is relevant for the storage database.
- Whisper and Ceres use the same config options.
- I'm not aware of any other storage backend for carbon-cache (other tsdb's implement the carbon protocol natively).

I chose the quick path for now, with intent to properly address it later.

> FYI, I'd like to hold off merging Ceres from the other branch until graphite-web supports reading past retentions which is something I've been working on getting put in

This is something I implemented in Ceres itself for now (I think you know this), I should rebase that PR for FYI purposes.  Good to know that someone is working on pushing it out to the client though, that is definitely the right way to go about it.

I'll raise a supplementary PR then which completes https://github.com/graphite-project/carbon/compare/graphite-project:master...iain-buclaw-sociomantic:tsdatabase2 but stops short of actual Ceres support.

PR #490, opened 2015-12-07
Cherry-pick FORWARD_ALL config option from 0.9.x branch

iain-buclaw-sociomantic:
Second change from 0.9.x branch pulled into master.  As for the third change, I'll leave the AMPQ things alone.  Best leave it for someone who  actually uses/understands it.


iain-buclaw-sociomantic:
Note, to pass the unittests (and to not change behaviour of master), I defaulted the ``FORWARD_ALL`` value to ``True`` instead of ``False`` as in 0.9.x

PR #491, opened 2015-12-08
Fix documentation on random write strategy

iain-buclaw-sociomantic:
The configuration option has always been ``random``, not ``naive``.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -129,7 +129,7 @@ LOG_CACHE_QUEUE_SORTS = False
 # there are a large number of metrics which receive very frequent updates OR if
 # disk i/o is very slow.
 #
-# naive - Metrics will be flushed from the cache to disk in an unordered
+# random - Metrics will be flushed from the cache to disk in an unordered
 # fashion. This strategy may be desirable in situations where the storage for
 # whisper files is solid state, CPU resources are very limited or deference to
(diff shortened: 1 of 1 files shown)


iain-buclaw-sociomantic:
Oh wait, ``random`` seems to be the dumber, slower version of ``naive``.  I'll close for now. :-)

drawks:
I think naive disappeared in master around 88a646 when @mleinart did the big merge of stuff from megacarbon. CACHE_WRITE_STRATEGY got rejiggered into the pipeline friendly DrainStrategy class in the cache module. Or at least that appears to be the intent here.

iain-buclaw-sociomantic:
It looks more like there are four strategies. ``sorted``, ``max``, ``random``, and ``whatever-name-you-give-me-I-will-be-used-naive``.  The difference between ``random`` and ``naive`` being that the former uses ``random.choice`` to select an item from keys() and the latter just pops the next iterkey from the dictionary.

drawks:
Yes, but it is in an entirely different part of the code as well.

iain-buclaw-sociomantic:
Is it? https://github.com/graphite-project/carbon/blob/d9349462c5ffa5a6d55fe9f2779e88eb44e5fd6a/lib/carbon/cache.py#L151-L160

Then again, I'm sure you know this code better than me.

PR #493, opened 2015-12-09
Port benchmark_cache.py from 0.9.x

iain-buclaw-sociomantic:
It strikes me that this is something that should be monitored and improved.  It seems that the only usable strategy (by an order of magnitude) when dealing with an upwards of a million metrics is ``sorted``.

At the very least ``naive`` should have a higher through-put as it should not iterate the data more than once (got another PR for this).


iain-buclaw-sociomantic:
Thanks.

PR #494, opened 2015-12-09
Implement NaiveStrategy class

iain-buclaw-sociomantic:
This is a hybrid between ``_MetricCache(None)`` and ``_MetricCache(SortedStrategy)``.

Like ``SortedStrategy``, it generates a queue to drain, however nothing special is done to sort the queue in any way.

When testing against the only available benchmark I could find for carbon (in 0.9.x branch), this made ``naive`` ever so slightly faster than the ``sorted`` strategy at least when draining in memory.  Haven't tested on any real SSDs or Disks yet, however.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -44,6 +44,23 @@ class DrainStrategy(object):
     raise NotImplemented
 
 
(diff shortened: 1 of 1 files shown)


drawks:
:+1:  rad

Issue #360, opened 2015-01-12
Current master occasionally produces errors


garrickp:
Today I updated our environment (consisting of graphite-web, carbon, and whisper packages from this project) to the current master, and we started getting errors on about every other request to graphite. Pasted below is the error from the django debug. The error appears to be in a recent commit in the carbon cache code, as I was able to roll back the carbon cache install from commit 01062f182ddc75081f16f53a674213059cc044dd to commit de072d869087f3bc75a046636dd7f73fafded5e8 and the problem went away. Graphite web and whisper remain on the current head as of the time of this message.

```
Environment:


Request Method: GET
Request URL: https://[...]/graphite/render/?target=ping.[...].bresnan&from=-6hours&format=json

Django Version: 1.5.11
Python Version: 2.7.3
Installed Applications:
('graphite.metrics',
 'graphite.render',
 'graphite.browser',
 'graphite.composer',
 'graphite.account',
 'graphite.dashboard',
 'graphite.whitelist',
 'graphite.events',
 'graphite.url_shortener',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.admin',
 'django.contrib.contenttypes',
 'django.contrib.staticfiles',
 'tagging')
Installed Middleware:
('graphite.middleware.LogExceptionsMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.gzip.GZipMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware')


Traceback:
 File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  113.                         response = callback(request, *callback_args, **callback_kwargs)
File "/opt/graphite/webapp/graphite/render/views.py" in renderView
  111.         seriesList = evaluateTarget(requestContext, target)
File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTarget
  7.   result = evaluateTokens(requestContext, tokens)
File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTokens
  18.     return evaluateTokens(requestContext, tokens.expression)
File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTokens
  21.     return fetchData(requestContext, tokens.pathExpression)
File "/opt/graphite/webapp/graphite/render/datalib.py" in fetchData
[...]

PR #496, opened 2015-12-10
Fix carbonlink returning a dict of {ts: val, ...} rather than a list of [(ts, val), ...]

mleinart:
This is the cause of #360, graphite-project/graphite-web#1122, graphite-project/graphite-web#1156, graphite-project/graphite-web#1139, et al

The workaround applied in #1122 doesn't require any changes to graphite-web for this to work (though it should probably be removed from there once this is merged

### DIFF
--- a/lib/carbon/protocols.py
+++ b/lib/carbon/protocols.py
@@ -143,7 +143,7 @@ class CacheManagementHandler(Int32StringReceiver):
     request = self.unpickler.loads(rawRequest)
     if request['type'] == 'cache-query':
       metric = request['metric']
-      datapoints = MetricCache.get(metric, [])
+      datapoints = MetricCache.get(metric, {}).items()
       result = dict(datapoints=datapoints)
       if settings.LOG_CACHE_HITS:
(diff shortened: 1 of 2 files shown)


toni-moreno:
Hi @mleinart , can this PR fix this other issue https://github.com/graphite-project/graphite-web/issues/1399.

I have already applied the workaround in #1122 , and still watching these: ''float' object is not iterable' errors.

I can see the errors , but only when querying carbon metrics and only when the requested carbon cache is heavily loaded ( perhaps blocked on I/O).

Could you help me to debug this error?

mleinart:
This is the only cause I can spot for these errors, though it's possible
there's another bug lurking here.

If you want to investigate a bit further before applying the carbon patch,
you can apply this patch from #1433 to graphite-web to reveal the traceback
of the actual exception causing the problem:
https://patch-diff.githubusercontent.com/raw/graphite-project/graphite-web/pull/1433.patch

That should let us narrow down what other code path is affected by
graphite-project/carbon#360 (if any)

PR #508, opened 2015-12-20
Remove remaining dependencies on whisper

iain-buclaw-sociomantic:
As per #484

This does the following:
- Log the metric path instead of the filesystem path in ``writeCachedDataPoints``, because the storage database may not have an underlying filesystem.
- Move ``parseRetentionDef`` into carbon (yes, this is pretty much duplicated from whisper).
- Add ``aggregateMethods`` and ``validateArchiveList``abstractions, which are not found in ``megacarbon`` branch (megacarbon chooses not to care).

With this, the only place where whisper is imported is in database.py.

### DIFF
--- a/lib/carbon/database.py
+++ b/lib/carbon/database.py
@@ -47,6 +47,10 @@ class TimeSeriesDatabase(object):
     "Return filesystem path for metric, defaults to None."
     pass
 
+  def validateArchiveList(self, archiveList):
+    "Validate that the database can handle the given archiveList."
+    pass
+
 
 try:
   import whisper
@@ -123,3 +127,9 @@ else:
     def getFilesystemPath(self, metric):
       metric_path = metric.replace('.', sep).lstrip(sep) + '.wsp'
       return join(self.data_dir, metric_path)
+
+    def validateArchiveList(self, archiveList):
+      try:
+        whisper.validateArchiveList(archiveList)
+      except whisper.InvalidConfiguration, e:
+        raise ValueError("%s" % e)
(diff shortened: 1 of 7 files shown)


iain-buclaw-sociomantic:
Rebased against master.

obfuscurity:
@iain-buclaw-sociomantic Is there a planned Whisper PR to remove the duplicated bits?

@cbowman0 Do these test changes look ok to you?

iain-buclaw-sociomantic:
@obfuscurity - no planned Whisper PR.  I'll have to check on whether it *can* be removed, or we'll have to stick with duplication because whisper uses `parseRetentionDef` internally.

iain-buclaw-sociomantic:
Looks like it's used by whisper tools:
https://github.com/graphite-project/whisper/blob/876382fb8acdbeef0a19695c23a9ea89271d224b/bin/whisper-create.py#L59

https://github.com/graphite-project/whisper/blob/876f2eb966a3f3a26923601cfc6495b9ae742e94/bin/whisper-resize.py#L69-L70

cbowman0:
The test changes look ok to me.

obfuscurity:
@iain-buclaw-sociomantic Sorry, I think my recent merge caused this to need a rebase. Would you mind?

P.S. You've been running this in production without issues?

iain-buclaw-sociomantic:
Yeah, I'm using the ceres storage backend in production, which needs this.

iain-buclaw-sociomantic:
Thanks!

PR #315, opened 2014-10-03
Bring master up to speed with 0.9.x

jssjr:
Hot on the heels of https://github.com/graphite-project/carbon/issues/312 is this PR which re-integrates 0.9.x into master. There were a few merge conflicts, but far less than I'd feared.

It would rock if someone else could take this branch and test it out some more. My initial tests (start a few caches, write data, verify data) all pass, but are rather simplistic.

PR #509, opened 2015-12-20
Merged PR #190

iain-buclaw-sociomantic:
Noticed just by chance in this PR https://github.com/graphite-project/carbon/pull/315/files#diff-db27808e2f8d53b9406df573ce15b714L413

Original commit message was 119 (https://github.com/graphite-project/carbon/commit/3bd7355b0c07fe94cf7b0728c129b378d838ed7a)

This should fix something, as the code logic seems odd, but haven't checked.

### DIFF
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -410,8 +410,8 @@ class CarbonRelayOptions(CarbonCacheOptions):
         settings["relay-rules"] = self["rules"]
 
         if self["aggregation-rules"] is None:
-            self["rules"] = join(settings["CONF_DIR"], settings['AGGREGATION_RULES'])
-        settings["aggregation-rules"] = self["rules"]
+            self["aggregation-rules"] = join(settings["CONF_DIR"], settings['AGGREGATION_RULES'])
+        settings["aggregation-rules"] = self["aggregation-rules"]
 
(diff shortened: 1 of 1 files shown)


obfuscurity:
I _think_ you're right but we really should test this to ensure we're not regressing a partial fix in https://github.com/graphite-project/carbon/pull/300. I'll try to take a closer look and run some tests this weekend.

mleinart:
I hope I remember this right but I believe what happened with this was that
I didn't see it at the time and manually went through looking for  commits
to 0.9.x missing in master and applied them, breaking the mergability of
this

I talked to jssrjr at the time (at Monitorama 2015) and apologized and I
think we agreed to drop it.

FWIW (with me not being active any longer) my opinion is that the branches
are so far diverged that this would A) not be worth the effort (it's
literally easier to diff the files in each checkout and verify the
differences are expected) and B) would create a huge mess in the commit
history.

In any case I doubt this PR is usable as-is, you probably need to start
fresh if you really want to do a merge.

Finally, If something like this is done it should be squash-merged in. The
commit hygiene in the graphite repos is already terrible (due to GitHub's
default merge strategy and our lack of commit standards for committers) and
inserting so many commits into the history with all of the merge-commits
necessary will make it  far worse.

mleinart:
Ah damn, I completely misinterpreted this as a few to merge that old pr :(

Carry on...

obfuscurity:
Confirmed this was also proposed as part of #315.

PR #510, opened 2015-12-22
fix MetricCache KeyError

shadow4125:
When I ran carbon-cache I can see the errors in console.log:
```
22/12/2015 08:48:26 :: Unhandled Error
  File "/home/q/graphite/lib/python2.7/site-packages/twisted/protocols/basic.py", line 755, in dataReceived
    self.stringReceived(packet)
  File "/opt/graphite/lib/carbon/protocols.py", line 126, in stringReceived
    self.metricReceived(metric, datapoint)
  File "/opt/graphite/lib/carbon/protocols.py", line 72, in metricReceived
    events.metricReceived(metric, datapoint)
  File "/opt/graphite/lib/carbon/events.py", line 20, in __call__
    handler(*args, **kwargs)
--- <exception caught here> ---
  File "/opt/graphite/lib/carbon/pipeline.py", line 25, in run_pipeline
    for out_metric, out_datapoint in processor.process(metric, datapoint):
  File "/opt/graphite/lib/carbon/cache.py", line 32, in process
    MetricCache.store(metric, datapoint)
  File "/opt/graphite/lib/carbon/cache.py", line 162, in store
    self[metric][timestamp] = value
exceptions.KeyError: 't.shadow.test'
```
I see store() setdefault dict first
```
self.setdefault(metric, {})
```
but, maybe this time pop()
```
datapoint_index = dict.pop(self, metric)
```
so I use defaultdict replace dict

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -15,6 +15,7 @@ limitations under the License."""
 import time
 from operator import itemgetter
 from random import choice
+from collections import defaultdict
 
 from carbon.conf import settings
 from carbon import events, log
@@ -99,16 +100,14 @@ class SortedStrategy(DrainStrategy):
     return self.queue.next()
 
 
-class _MetricCache(dict):
+class _MetricCache(defaultdict):
   """A Singleton dictionary of metric names and lists of their datapoints"""
   def __init__(self, strategy=None):
     self.size = 0
     self.strategy = None
     if strategy:
       self.strategy = strategy(self)
-
-  def __setitem__(self, key, value):
-    raise TypeError("Use store() method instead!")
+    super(_MetricCache, self).__init__(dict)
 
   @property
   def counts(self):
@@ -143,14 +142,13 @@ class _MetricCache(dict):
     return sorted(self.get(metric, {}).items(), key=by_timestamp)
 
   def pop(self, metric):
-    datapoint_index = dict.pop(self, metric)
+    datapoint_index = defaultdict.pop(self, metric)
     self.size -= len(datapoint_index)
     self._check_available_space()
 
     return sorted(datapoint_index.items(), key=by_timestamp)
 
   def store(self, metric, datapoint):
-    self.setdefault(metric, {})
     timestamp, value = datapoint
     if timestamp not in self[metric]:
       # Not a duplicate, hence process if cache is not full


iain-buclaw-sociomantic:
I'm seeing this too.

mleinart:
This looks like a race condition to me: a pop is happening sometime after the setdefault but before the datapoint is actually stored. It turns out setdefault() behaves like a check-and-set, i.e. the pop() will remove what it did (if anything).

I have an idea to fix the race, but I don't see a reason against just using a defaultdict here so I think I'm :+1: on this

iain-buclaw-sociomantic:
Should probably do something like in the 0.9.x branch? https://github.com/graphite-project/carbon/blob/ae1464d02728e22298ff480296410caae381d025/lib/carbon/cache.py#L18-L21

deniszh:
@iain-buclaw-sociomantic 
IIRC that's a workaround for python 2.6 and we're pointing to 2.7 compatibility to master. So, up to you. Do not hurt but do not need.

mleinart:
IIRC we're still nursing along 2.4 support in 0.9.x. There's a pure python backport of defaultdict in carbon.util in 0.9.x branch but it's not any more threadsafe than the current way of doing it. One option for 0.9.x would be to re-add locks that were taken out in fe7686f4097dc9c81bb75f001b1cdde8d29fa251.

I'm surprised that we've gone so long without seeing reports of this though. @iain-buclaw-sociomantic @shadow4125 how often are you seeing this and what version of Python are you on?

iain-buclaw-sociomantic:
@mleinart - This is the first time I encountered it, but it happened twice on two isolated servers running carbon-cache within the space of a week of each other - both having an uptime of no more than 10 days.

Some important things to note about my environment which have likely lead to this:
- A couple of weeks back we sorted out a problem where we were handling 4x more data than we should have been receiving.  So they are now no longer under heavy strain (writing out 180K metrics every 30 seconds).
- I am using the ``naive`` strategy I implemented, which admittedly gives this a higher chance to trigger compared to using ``sorted`` strategy.
- For me, this has only ever fired when the instrumentation service has sent a metric to cache.

I am running PyPy 4.0.1.

iain-buclaw-sociomantic:
Saying that, when we were running handling 1M metrics per carbon-cache node, I would notice that the ``MetricCache.size`` would become a negative number after a few days of running.  Maybe it's a different issue altogether though (which I can see a likely fix for in that commit you linked):
```
+  @property
+  def size(self):
+    return reduce(lambda x, y: x + len(y), self.values(), 0)
```

iain-buclaw-sociomantic:
>  Maybe it's a different issue altogether though (which I can see a likely fix for in that commit you linked)

Actually, applying that would cause a terrible performance regression if the cache were to grow very large.  Ignore me here. :-)

mleinart:
Merging this since defaultdict is more clear here, fixes the immediate symptom, and doesnt seem to perform any worse. I'd also like to bring attention to #527 which should prevent any further races in this part of the code, including negative cache sizes mentioned above

iain-buclaw-sociomantic:
Thanks, I'll give it a try out.

PR #513, opened 2015-12-30
Add CACHE_LOW_WATERMARK_PCT setting.

iain-buclaw-sociomantic:
Gives user control of what percentage the CACHE_SIZE_LOW_WATERMARK value
should be of MAX_CACHE_SIZE, which in turn allows to fine tuning when
cacheSpaceAvailable should be triggered if USE_FLOW_CONTROL=True.

In a site where there is a high volume of metrics incoming (500k/minute)
having a higher percentage could result in the carbon-cache service stalling
when the drain throughput falls as it will always hitting cacheTooFull every
couple of seconds.  Reducing the percentage gives it more time to recover.


iain-buclaw-sociomantic:
Actually, the problem I am seeing was something entirely different.  The cache was not recovering because something happened to cause hundreds of TCP sockets to open between the relay and cache, which slowed the entire cache service down to a crawl.

It appears that while ``USE_FLOW_CONTROL`` does block the current connection from receiving more metrics, it doesn't prevent the relay from opening a new connection to try the same thing again (and fail).

iain-buclaw-sociomantic:
I'll leave this open though, as it may be useful to have this configurable, rather than hard coded.

PR #515, opened 2016-01-08
Add MAX_RECEIVER_CONNECTIONS setting.

iain-buclaw-sociomantic:
Limits the number of 'connectedMetricReceiverProtocols' that a given
receiver can handle at any given time.

This is to prevent, for example, a carbon-relay opening up thousands of
connections to a carbon-cache service, both reducing the throughput of
the service and causing out of TCP memory errors.

Should also fix the observed behaviour I was trying to resolve in #513


iain-buclaw-sociomantic:
Holy cow, how on earth did I not see the ``METRIC_CLIENT_IDLE_TIMEOUT`` setting.  Surely this should have a value other than ``None``.

OK, removing first commit, but keeping the latter as it was useful for diagnosis.

iain-buclaw-sociomantic:
Actually, no.  I've reverted my mind, I'll keep this as is for now.

mleinart:
Awesome

iain-buclaw-sociomantic:
Thanks.

Issue #497, opened 2015-12-11
carbon-cache  strange behaviour: Mb reads from disk  greater than Mb writes (without graphite-web render hits)


toni-moreno:
Hi @mleinart , @SEJeff 

I'he been working in a big performance issue when working with graphite.

* graphite-web ( https://github.com/graphite-project/graphite-web/tree/8710ff76f0b4571481d295d997a83e4c895dd092 )
* carbon ( b80ce915a5e420b46e6972512801491e536db1b6 )
* whisper (https://github.com/graphite-project/whisper/tree/1e96c0cd1dc0b361177c585033cfbbb5711a191f)

As you can see there is much more load from reads than  load from writes over the metric dedicated disks (even with only 10 hits in the graphite , in the meanwhile carbon-caches are storing 170K metrics/minute )

![image](https://cloud.githubusercontent.com/assets/5883405/11741616/68c093a8-9ffa-11e5-8418-f097aa4eb3fa.png)

![image](https://cloud.githubusercontent.com/assets/5883405/11741765/53b9f52a-9ffb-11e5-86b3-fb8aea8a123c.png)

You can see each Read/Write with iotop ( now at 12:00 aprox)
````
Total DISK READ: 90.27 M/s | Total DISK WRITE: 0.00 B/s
  PID  PRIO  USER     DISK READ  DISK WRITE  SWAPIN     IO>    COMMAND
 2322 be/4 carbon     13.61 M/s    3.42 M/s  0.00 % 40.25 % python bin/carbon-cache.py --instance=wr07c start
 2302 be/4 carbon     14.46 M/s    4.03 M/s  0.00 % 38.97 % python bin/carbon-cache.py --instance=wr07a start
 2312 be/4 carbon     14.79 M/s    4.03 M/s  0.00 % 38.65 % python bin/carbon-cache.py --instance=wr07b start
 2352 be/4 carbon     15.95 M/s    3.62 M/s  0.00 % 37.89 % python bin/carbon-cache.py --instance=wr07d start
61286 be/4 carbon     15.05 M/s    4.17 M/s  0.00 % 37.53 % python bin/carbon-cache.py --instance=wr07f start
61276 be/4 carbon     16.62 M/s    3.66 M/s  0.00 % 37.49 % python bin/carbon-cache.py --instance=wr07e start
```

These estrange reads  over disk are impacting on the CPU  ( you could see less io wait when Reads decrease)

[...]

PR #523, opened 2016-02-18
Enable whisper caching

pu239ppy:
Exposes cache enablement and age limit through settings
Launches a looping task in order to age items out of a cache
relies on https://github.com/graphite-project/whisper/pull/157


bmhatfield:
As a devils advocate, I'd advise that trying to cache these headers in a per-process way (for each carbon as well as every python/django process) is very inefficient (on my system, I would end up with about 10 copies of the same "cache") and leads to worse overall performance than if we left this memory unused by python and let the filesystem cache manage them instead.

See also: https://github.com/graphite-project/carbon/issues/497#issuecomment-163949580

pu239ppy:
Brian, this setting is not mandatory, and in fact disabled by default.
While it would be possible to have a unified cache with a caching system I
am reluctant to add more components (caching layer such as redis) to
carbon.

bmhatfield:
Thanks for the additional context. I agree that disabled by default, if we go forward, is a good approach for introducing new behaviors such as this.

I guess my statement could be restated in the form of a question: Why add this, or a common caching layer like Redis (which already exists in Graphite as memcached), when the "do nothing and let the Linux filesystem cache handle it" is the most efficient and performant outcome?

Anyways, if other folks see value in this and want to move forward, I don't need to block this change further.

pu239ppy:
If I recall correctly memcache only exists on graphite web, carbon does not reply on external services.  The filesystem buffers to which I assume you refer as as the "do nothing and let the Linux filesystem cache handle it" are not deterministic and are not guaranteed to keep the data I need.  Although it would be interesting to see what hits/misses I encounter (would have to figure out how to instrument that)

obfuscurity:
Yes, but Whisper also relies on Graphite-Web to read data. Well, technically anything can use the Whisper library, but is that the use case (non-Graphite-Web readers) this is intended to address?

pu239ppy:
These particular enhancement work in concert *only* to improve the performance of metric write operations.  Currently a write operation entails a read in to retrieve the whisper header.  This read is what I want to eliminate, in fact in whisper.py someone already took the steps toward that, I am merely enhancing this process by introducing per item expiration to the cache entry and a looping task to manage that expiry.

iksaif:
I had to enable header caching once on carbon before we disabled buffering on reads in whisper. It helped a little bit reducing the load, not much though.

On biggraphite (https://github.com/criteo/biggraphite) we ended up using LMDB to share the cache across multiple processes. memcached was simply too slow for writes: you can't spend 1ms each time you write a point.

piotr1212:
fwiw: I once tested performance gains with CACHE_HEADER = True hardcoded in whisper.py on  a production system (> 2.000.000 unique metrics) and didn't found any noticable performance improvements. I assumed that the file is read anyway for aggregation and the header got read from fs-cache, not causing any additional io to the disks.

this might have changed since the buffering patch:  https://github.com/graphite-project/whisper/commit/a3efa8453df65554e3636a32dbf39a1a4032993f

iksaif:
CACHE_HEADER = True will only help if you are low on memory for the buffer cache. But you're right, most of the time this is cached by the os anyway.

PR #527, opened 2016-03-09
Synchronize calls to MetricCache.drain_metric from writer thread

mleinart:
The writer thread drains from the cache in a separate thread from the reactor. This sometimes leads to weird race conditions (e.g. negative cache size metrics). Instead, run the cache drain on the main reactor
thread (the place where the cache is written to).

This keeps the rest of the draining logic (cache sorting, etc) in the writer thread so as not to block the event loop.

I've done some basic performance testing on this, the expectation being that there shouldn't be much difference. The comparisons I did show no difference in throughput but CPU seemed to be a little more consistent (same moving avg, smaller range).

I'd love if someone could canary this in production and report back

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -16,6 +16,10 @@ import time
 from operator import itemgetter
 from random import choice
 from collections import defaultdict
+from threading import current_thread
+
+from twisted.internet import reactor
+from twisted.internet.threads import blockingCallFromThread
 
 from carbon.conf import settings
 from carbon import events, log
@@ -142,6 +146,11 @@ class _MetricCache(defaultdict):
     return sorted(self.get(metric, {}).items(), key=by_timestamp)
 
   def pop(self, metric):
+    if current_thread().name == 'MainThread':
+      return self._pop(metric)
+    return blockingCallFromThread(reactor, self._pop, metric)
+
+  def _pop(self, metric):
     datapoint_index = defaultdict.pop(self, metric)
     self.size -= len(datapoint_index)
     self._check_available_space()
--- a/lib/carbon/service.py
+++ b/lib/carbon/service.py
@@ -20,12 +20,10 @@ from twisted.internet.protocol import ServerFactory
 from twisted.python.components import Componentized
 from twisted.python.log import ILogObserver
 # Attaching modules to the global state module simplifies import order hassles
-from carbon import state, events, instrumentation, util
+from carbon import state, util
 from carbon.exceptions import CarbonConfigException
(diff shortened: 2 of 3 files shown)


iain-buclaw-sociomantic:
I did give this a try, however the carbon-cache service failed to start.

mleinart:
@iain-buclaw-sociomantic Do you have an error you can paste?

iain-buclaw-sociomantic:
```
  File "/srv/graphite/bin/carbon-cache.py", line 32, in <module>
    run_twistd_plugin(__file__)
  File "/srv/graphite/site-packages/carbon/util.py", line 96, in run_twistd_plugin
    config.parseOptions(twistd_options)
  File "/srv/graphite/site-packages/twisted/application/app.py", line 579, in parseOptions
    usage.Options.parseOptions(self, options)
  File "/srv/graphite/site-packages/twisted/python/usage.py", line 255, in parseOptions
    self._dispatch[optMangled](optMangled, arg)
  File "/srv/graphite/site-packages/twisted/python/usage.py", line 411, in <lambda>
    fn = lambda name, value, m=method: m(value)
  File "/srv/graphite/site-packages/twisted/application/app.py", line 496, in opt_reactor
    raise usage.UsageError(msg)
UsageError: The specified reactor cannot be used, failed with error: reactor already installed.
See the list of available reactors with --help-reactors
```

iain-buclaw-sociomantic:
Doesn't happen if I comment out from `lib/carbon/cache.py`
```
from twisted.internet import reactor
```

mleinart:
@iain-buclaw-sociomantic Finally got around to fixing the startup if you'd like to try. The issue was with the reactor import - the Twisted reactor cant be imported until service start and adding the import to cache.py caused it to be imported through instrumentation.py through service.py when the service is first loaded, pre-start.
This fix for the import issue is IMO neither terrible nor great - ultimately we need to at some point reorganize to avoid circular imports and (maybe?) memoize the reactor. In the meantime, this keeps things running

obfuscurity:
@mleinart This is an alternative to #551? Does anyone have time to benchmark these different approaches?

mleinart:
Finally did some relatively thorough performance testing on this and... We shouldn't merge it.

While this did not appear to have any negative effect on my (modest) production install, when I simulated the worst-case scenario (tons of simultaneous updates from different agents with a high MAX_UPDATES_PER_SEC) CPU usage was significantly higher.

In contrast, in performing the same tests with #551, I found no observable difference in performance. These results are completely contrary to my expectation but.. ¯\_(ツ)_/¯ the numbers dont lie

PR #531, opened 2016-03-11
Update carbon.conf.example

piotr1212:
there is no [carbon] section, there only are cache|relay|aggregator sections


piotr1212:
reported by danfrincu on irc

gwaldo:
Yup, that's a pretty glaring mistake... Thanks!

PR #535, opened 2016-03-21
Support FADVISE_RANDOM to avoid disk thrashing

yuuki:
- On our system, it occurs heavy disk thrashing (https://en.wikipedia.org/wiki/Thrashing_(computer_science)), and the disk i/o activity is very high load.
- We think it is bacause that a large amount of whisper write(2)'s pageback drives out page cache to disk.
- Avoid file readahead by posix_fadvise(2) with POSIX_FADV_RANDOM to solve disk thrashing

I've written a test (gist here: https://gist.github.com/yuuki/8d5d386115b0f01b5371), which runs whisper_update_many 100 datapoints by one datapoint on 100 whisper files.

```
% sync; echo 3 | sudo tee /proc/sys/vm/drop_caches ; python bin/whisper-update-many-fadvice-test.py n | grep read
read_bytes: 20877312
```

```
% sync; echo 3 | sudo tee /proc/sys/vm/drop_caches ; python bin/whisper-update-many-fadvice-test.py f | grep read
read_bytes: 9961472
```

FADVISE_RANDOM version reduced read_bytes from disk by half.

PR for whisper is here https://github.com/graphite-project/whisper/pull/162.


deniszh:
So, POSIX_FADV_RANDOM just disables readahead, but on a file level, right? Nice catch, then! 
:+1:

lib/carbon/database.py
```diff
@@ -77,6 +77,13 @@ def __init__(self, settings):
         else:
           log.err("WHISPER_LOCK_WRITES is enabled but import of fcntl module failed.")
 
+      if settings.WHISPER_FADVISE_RANDOM:
+        if whisper.CAN_FADVISE:
```
deniszh: Maybe we need to add try-catch block here for ignoring of non-compatible whisper lib?

yuuki:
> POSIX_FADV_RANDOM just disables readahead, but on a file level, right?

That's exactly right!

lib/carbon/database.py
yuuki: I think so too, I will fix it.

yuuki:
I added the try-catch block. If whisper is not compatible, carbon ignores and just log a message.

yuuki:
@deniszh Thank you for your advice. Is there any other issue?

This patch actually reduces dramatically inactive memory on our production. And resolved disk threshing.

before `/proc/meminfo`
```
Active(file):    5387160 kB
Inactive(file): 37566804 kB
```

after `/proc/meminfo`
```
Active(file):   32252136 kB
Inactive(file):  7231020 kB
```

I'm working in Mackerel team (ref https://github.com/graphite-project/graphite-web/pull/1447#issuecomment-171148785). 

We hope this pull request to be accepted.

Songmu:
@deniszh
How about it?

deniszh:
LGTM! 
I can merge it but - @obfuscurity @mleinart @bmhatfield - could you guys also check? Thanks!

bmhatfield:
This looks good to go to me. Obviously we'd also need to merge https://github.com/graphite-project/whisper/pull/162 for this to realize any value, but I like that it's backwards compatible across whisper versions and off by default.

I'm not 100% on the real-world performance effects of this change; though it looks reasonable enough that if it helps people tune some systems, I think it's better than not having it.

yuuki:
@obfuscurity @mleinart Do you have any problems with this PR? (https://github.com/graphite-project/whisper/pull/162 has already merged.)

obfuscurity:
👍

yuuki:
@deniszh Thanks for your help. Some graphite project members given this PR LGTM, so I think there are no longer a problem for merging it. How about it?

yuuki:
Thanks!

PR #540, opened 2016-03-28
Change MetricLineReceiver delimiter to tab

bdrich:
This should resolve common errors with the newline character. 
For example, in carbon console.log:
```
[listener] invalid line () received from client <HOST>:<PORT>, ignoring
```


drawks:
:-1: One delimiter is as good as the other, changing the line protocol in a way that breaks backwards compatibility driven by a questionably justified change doesn't sound like a good change.

PR #548, opened 2016-04-28
Move PYTHONPATH setting to where $GRAPHITE_DIR is even defined

jblaine:
Fixes reference to GRAPHITE_DIR before it's even been set.


gwaldo:
LGTM.  Do any other @graphite-project/committers have an opinion on this?

bitprophet:
I haven't used this particular file in my installations but this sure seems like a straightforward fix, :+1:

PR #550, opened 2016-04-29
Check for rename from external log rotation

jonsulman:
The log rotation support checks if a log file has disappeared, but doesn't detect log files that have moved if the log file path has been recreated. This adds a check to see if the inode of the currently open log file matches the inode of the file path, and reopens the log file if it does not match.


obfuscurity:
LGTM, thanks! 🍰

Issue #528, opened 2016-03-09
[carbon-cache] MetricCache.size reported as being below zero


iain-buclaw-sociomantic:
After a day or so of running, it is not uncommon for the reported number of datapoints held by `MetricCache` to drop below zero.

![metriccache_size-bug](https://cloud.githubusercontent.com/assets/11678319/13633367/de9407ba-e5ee-11e5-811d-64b37d7de6be.png)

Filtering out the high spikes (these are when ceres-maintenance is ran overnight to roll up data), it looks like it will keep on dropping to a lower and lower value.

![metriccache_size-bug2](https://cloud.githubusercontent.com/assets/11678319/13633472/863f6cfc-e5ef-11e5-8af8-57955e1a507b.png)

This would affect setting `MAX_CACHE_SIZE`, as you can no guarantee that it will be adhered to.

The only places where the size is modified is in `MetricCache.pop` and `MetricCache.store`, what it likely indicates is that the `+=` and `-=` operations aren't atomic in Python.  But I'm at loss over how it could go out of sync negatively.  The increment and decrement should balance each other out no matter what order they are done in.  I can only take out of this that either:
- Something is bypassing `MetricCache.store` and pushing more datapoints into cache.
- The reported length of the pop'd datapoints somehow returns a wrong number.

I'll give #527 a try, as it should deal with possible synchonization issues.  However I think that it should also guard `MetricCache.store` from being called at the same time from multiple threads too.

PR #551, opened 2016-05-02
Lock writes to MetricCache size and dictionary

iain-buclaw-sociomantic:
My attempt at trying to address #528.

I don't really like the idea of it, as it slows down the cache.  However this was the best case method I could find.

Using this as a base benchmark:
```
Benchmarking single metric MetricCache store...
    datapoints: 1000       msecs: 1
    datapoints: 10000      msecs: 19
    datapoints: 100000     msecs: 174
    datapoints: 1000000     secs: 1.71134
Benchmarking unique metric MetricCache store...
    datapoints: 1000       msecs: 3
    datapoints: 10000      msecs: 33
    datapoints: 100000     msecs: 321
    datapoints: 1000000     secs: 3.2306
Benchmarking single metric MetricCache drain...
CACHE_WRITE_STRATEGY: max
    datapoints: 1000       usecs: 202
    datapoints: 10000      msecs: 1
    datapoints: 100000     msecs: 31
    datapoints: 1000000    msecs: 319
CACHE_WRITE_STRATEGY: naive
    datapoints: 1000       usecs: 168
    datapoints: 10000      msecs: 1
    datapoints: 100000     msecs: 31
    datapoints: 1000000    msecs: 318
CACHE_WRITE_STRATEGY: random
    datapoints: 1000       usecs: 174
    datapoints: 10000      msecs: 1
    datapoints: 100000     msecs: 32
    datapoints: 1000000    msecs: 319
CACHE_WRITE_STRATEGY: sorted
    datapoints: 1000       usecs: 313
    datapoints: 10000      msecs: 2
    datapoints: 100000     msecs: 31
    datapoints: 1000000    msecs: 318
Benchmarking unique metric MetricCache drain...
CACHE_WRITE_STRATEGY: max
    datapoints: 1000       msecs: 2
    datapoints: 10000      msecs: 27
    datapoints: 100000     [skipped]
    datapoints: 1000000    [skipped]
CACHE_WRITE_STRATEGY: naive
    datapoints: 1000       msecs: 2
    datapoints: 10000      msecs: 27
    datapoints: 100000     msecs: 277
    datapoints: 1000000     secs: 2.87606
CACHE_WRITE_STRATEGY: random
    datapoints: 1000       msecs: 2
    datapoints: 10000      msecs: 28
    datapoints: 100000     msecs: 279
    datapoints: 1000000    [skipped]
CACHE_WRITE_STRATEGY: sorted
    datapoints: 1000       msecs: 2
    datapoints: 10000      msecs: 28
    datapoints: 100000     msecs: 280
    datapoints: 1000000     secs: 2.88072
```
With this change, there was a 20% slowdown, with #527 there was a 40% slowdown.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 
 import time
+import threading
 from operator import itemgetter
 from random import choice
 from collections import defaultdict
@@ -103,6 +104,7 @@ class SortedStrategy(DrainStrategy):
 class _MetricCache(defaultdict):
   """A Singleton dictionary of metric names and lists of their datapoints"""
   def __init__(self, strategy=None):
+    self.lock = threading.Lock()
     self.size = 0
     self.strategy = None
     if strategy:
@@ -142,8 +144,9 @@ class _MetricCache(defaultdict):
     return sorted(self.get(metric, {}).items(), key=by_timestamp)
 
   def pop(self, metric):
-    datapoint_index = defaultdict.pop(self, metric)
-    self.size -= len(datapoint_index)
+    with self.lock:
+      datapoint_index = defaultdict.pop(self, metric)
+      self.size -= len(datapoint_index)
     self._check_available_space()
 
     return sorted(datapoint_index.items(), key=by_timestamp)
@@ -156,8 +159,9 @@ class _MetricCache(defaultdict):
         log.msg("MetricCache is full: self.size=%d" % self.size)
         events.cacheFull()
       else:
-        self.size += 1
-        self[metric][timestamp] = value
+        with self.lock:
+          self.size += 1
+          self[metric][timestamp] = value
     else:
       # Updating a duplicate does not increase the cache size
       self[metric][timestamp] = value


obfuscurity:
@iain-buclaw-sociomantic But the $64M question: does it fix the size reporting bug?

deniszh:
20%  slowdown just for reporting cache size? O_o

iain-buclaw-sociomantic:
> But the $64M question: does it fix the size reporting bug?

Graphs of carbon caches from the last 13 weeks.

carbon-cache

Never drops below zero.

> 20% slowdown just for reporting cache size? O_o

20% slowdown in draining the cache to zero in a serialized fashion, the benchmark itself it not very representative of real-life workloads where you have push and pops going in simultaneously.

But the meaning would be where it took `280ms` to drain 100,000 metrics, it now takes `320ms`.  Or `2.8 secs` to drain 1,000,000 metrics now takes `3.2 secs`.  I don't think this is a real problem, or at least if like me you only deal with an incoming 10K metrics every second.

deniszh:
Ah, good, thanks for explanation, @iain-buclaw-sociomantic !

iain-buclaw-sociomantic:
FYI, to explain that rise in cache towards the end, looks like I hit 100% CPU after the number of incoming metrics rose to 20K metrics a second using this patch.  YMMV, but I guess that could be an indicator of how much one carbon daemon with this patch can take before you need to start adding more.

iain-buclaw-sociomantic:
That was quick @obfuscurity.  Shouldn't this patch be scrutinized? :smiley:

obfuscurity:
Quick?

obfuscurity:
@iain-buclaw-sociomantic The patch itself looked reasonable to me and you've presented sane benchmark numbers. @mleinart scrutinized it a fair amount already and expressed to me his preference for this one over #527. Is there anything else you're concerned with, i.e. reason to revert this?

iain-buclaw-sociomantic:
Not from me.  I've been running this in production pretty much since this May.  I am more curious if there are better alternatives to a global thread lock.

mleinart:
Also of note, this same locking strategy was in use for many years
until fe7686f40 removed it.

My best guess as to why it doesnt seem to have much if any impact is that
we're only sharing these cache operations between two threads. As I
understand it, when a locked Lock is encountered, Python will context
switch to another thread in the hopes that that thread either wont be
waiting on it or will have it and free it up. In our case, that's true 100%
of the time - if the reactor thread encounters it, the writer thread is
locking it and needs to continue and vice versa.

I agree a lock like this seems brute force, but the "right way" to do it in
Twisted land as I understand it is the approach in #527 where anything
requiring access to the shared state is run on the reactor thread. Since
that seems to have a significant performance penalty (in one set of tests I
measured 60% Carbon cpu usage with that strategy vs 20% baseline), this
seems our best option at the moment

obfuscurity:
I haven't seen anything better yet. I'd still rather have it reported accurately with a "slight" performance penalty than emit bad stats. Hopefully we can iterate on this but we should definitely get it fixed in-tree for the time being.

obfuscurity:
Follow-up bit of info. I ran another round of benchmarks on some very fast NVME flash drives (two of them striped with LVM) to get some anecdotal comparisons before and after this change. There was very little visible change in performance with the exception of system load **decreasing** after this change. Fascinating.

In each of these graphs, the first test (1430-1520) was running d4a6a83536acb00cc31c00fe642aa20e6dbf8f7d. The second test (1530-) was running dfc1d75969cc4fe4309207e80ea4cf3c5e03c977.

screen shot 2016-09-08 at 12 29 52 pm
screen shot 2016-09-08 at 12 29 57 pm
screen shot 2016-09-08 at 12 30 04 pm
screen shot 2016-09-08 at 12 30 10 pm
screen shot 2016-09-08 at 12 30 20 pm
screen shot 2016-09-08 at 12 30 37 pm
screen shot 2016-09-08 at 12 30 42 pm

obfuscurity:
In case anyone's curious I was using https://github.com/gorsuch/haggar (on a separate host) to generate the benchmark data thusly:

```
GOPATH=~/gocode ./gocode/bin/haggar -agents=200 -metrics=4000 -carbon="x.x.x.x:2003"
```

Issue #557, opened 2016-05-30
[relay] and [aggregator] sections do not inherit from [cache] section


danfrincu:
Perhaps this is by design, but working with 0.9.15 on Centos 7 I've found that if you don't specify a USER in [relay] or [aggregator] sections it will start the processes under root, whereas for the [cache] section which has a default USER it will start the process under the said user, in this case carbon.

Given this behaviour, what other variables from the [cache] section are and are not inherited by the [relay] and [aggregator] sections?

I'm asking this because based on the docs I was expecting that the variables would be inherited from the [cache] section but I may have misinterpreted this and there is need for duplication of certain variables to the other sections, but it's unclear which are and which are not inherited, henceforth the question above.

Kind regards,
Dan

deniszh:
IIRC and it's by design, indeed. And none of the variables are inherited BETWEEN sections - only from [x] to [x:instance], but not between [x] and [y]...

danfrincu:
Is there a list of variables that exist in [cache] that can be applied to the other sections (not instances). The example being the USER variable, which is defined in [cache] but can be used in other sections.

deniszh:
Sorry, but why that list could be needed if there's no inheritance between different section exists?...

danfrincu:
Because for certain settings such as MAX_QUEUE_SIZE the config calls out the setting explicitly both in the [relay] and [aggregator] section, CARBON_METRIC_PREFIX is called out in all 3 sections, but the example above, USER, is only referenced in the [cache] section. 

Based on the example config file, the assumption is if the setting is only in one section, that's the only section it should be set to, but that's not entirely true as you can set USER in all the sections.

And for example, if I define STORAGE_DIR in [cache] as /something but I don't define it in the other sections, I assume that the other sections would use the default value for STORAGE_DIR and point someplace else.

A list of "what's available per section" is useful to deduct if any other settings can and should be added to the other sections.

deniszh:
STORAGE_DIR has no sense for [relay] and [aggregator], isn't it?
IMO if you put all combinations of all options to all sections config example will be "overloaded". 
Sorry, still see no much gain from the effort. But feel free to fix documentation or config example if you want to - https://github.com/graphite-project/graphite-web/tree/master/docs

danfrincu:
Yes, STORAGE_DIR wasn't the best example but you understand where I was going with this.

I understand adding all combinations to all sections would not be ideal.

[...]

PR #558, opened 2016-05-30
Adding USER to all config sections

danfrincu:
The USER setting is not inherited across sections, if this setting is not duplicated to [relay] and/or [aggregator] the carbon-cache will start under USER but the carbon-relay and carbon-aggregator ones will not. Adding this setting in the default example allows for understanding this option can be duplicated across sections.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -39,7 +39,7 @@ DATABASE = whisper
 ENABLE_LOGROTATION = True
 
 # Specify the user to drop privileges to
-# If this is blank carbon runs as the user that invokes it
+# If this is blank carbon-cache runs as the user that invokes it
 # This user must have write access to the local data directory
 USER =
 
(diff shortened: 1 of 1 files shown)


conf/carbon.conf.example
```diff
@@ -311,6 +311,11 @@ MAX_DATAPOINTS_PER_MESSAGE = 500
 # thousands of TCP connections reduce the throughput of the service.
 #MAX_RECEIVER_CONNECTIONS = inf
 
+# Specify the user to drop privileges to
+# If this is blank carbon-relay runs as the user that invokes it
+# This user must have write access to the local data directory
```
obfuscurity: This comment is inaccurate and unnecessary. Only carbon-cache requires write permissions to the data directory since it's the only daemon type that actually writes to Whisper. Would you mind removing this line and the one below for aggregator?

danfrincu: It was not intentional, I copied the entire paragraph initially. I've removed the lines you've mentioned and amended the change.

obfuscurity:
LGTM, thanks! 🌟

obfuscurity:
refs #557

PR #564, opened 2016-06-22
Don't stomp on config vaules

johnseekins:
With this change, adding a prefix or install-lib path in setup.cfg won't get overwritten.


deniszh:
:+1:
LGTM!

obfuscurity:
Nice, thanks! 🍰 ✨

PR #569, opened 2016-07-07
Fix missing imports

johnseekins:


obfuscurity:
Actually, I think this is a noop since `load_module` isn't used by anything. Originally @cdavis [introduced it](https://github.com/graphite-project/carbon/commit/6a6b943ba93d6dd70442d522d2f6f06a5c77169b) as an alternative to `imp.load_module` in the megacarbon branch. It then snuck into master with @mleinart's big [megacarbon refactor-merge](https://github.com/graphite-project/carbon/commit/88a646441cabf7def6721546c62ef19e5a1c7e1e).

@mleinart would you agree that we can just :boom: `load_module` here?

PR #573, opened 2016-08-23
Add service_identity requirement

obfuscurity:
Without it I'm seeing warnings every time I start or stop Carbon/Twisted.


obfuscurity:
Ugh, I was hoping this was an easy add. Looks like we have other secondary dep version issues. Sigh.

PR #580, opened 2016-08-30

obfuscurity:
Sorry for the PR spam, need to recreate this one again after a merge conflict (my git fu is just _terribad_). Deprecates the work in #437 which in turn replaced #231. I'd like to find a better var than `mdpu`, looking for suggestions.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -72,6 +72,21 @@ MAX_UPDATES_PER_SECOND = 500
 # cause carbon to create the files quickly but at the risk of increased I/O.
 MAX_CREATES_PER_MINUTE = 50
 
+# The datapoints for a metric would be written to disk only if the number of
+# datapoints in memory for that metric is greater than or equal to MIN_DATAPOINTS_PER_UPDATE. 
+# However this constraint is lifted if the cache size grows beyond LOW_CACHE_SIZE.
+# The default value for MIN_DATAPOINTS_PER_UPDATE is 1. Increase it to get a more
+# predictable write pattern across multiple carbon-cache instances, i.e. every
+# carbon-cache instance getting a share of disk I/O.
+#
+# It helps to avoid a large number of small disk-writes as well, making it a
+# suitable complement for the "random" CACHE_WRITE_STRATEGY.
+#
+# The default value for LOW_CACHE_SIZE is 80% of MAX_CACHE_SIZE.
+# Value for LOW_CACHE_SIZE must be less than 0.95 * MAX_CACHE_SIZE (which is the threshold for applying flow_control if enabled)  
+MIN_DATAPOINTS_PER_UPDATE = 1
+LOW_CACHE_SIZE = inf
+
 # Set the interface and port for the line (plain text) listener.  Setting the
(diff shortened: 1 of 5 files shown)


obfuscurity:
There's something broken with this change. Caches are neither able to create new files or write to existing ones. I'll try to look into this closer as time permits but I've removed it from the 1.0.0 milestone for now.

obfuscurity:
@pratX I know it's been a long time coming but if you have time to look at this it would be greatly appreciated.

pratX:
I will look into this.

iksaif:
Looked briefly at this because I want something similar. The idea is interesting but it looks like the implementation could use some love

Exocomp:
Did this feature make it to production?

deniszh:
@Exocomp - unfortunately not. I rebased it in #692 but not sure is it worth inclusion.
If you interested in this change - please test how it works.
Please also check timesorted writing strategy - it's a bit different, but maybe it's better fix then MDPU.

deniszh:
Closed in favor of #692

Exocomp:
@deniszh would the idea be to use MIN_TIMESTAMP_LAG with timesorted strategy ?

deniszh:
@Exocomp - I'm not sure what's problem are you trying to solve, probably MIN_TIMESTAMP_LAG is a solution, probably not. What are you trying to achieve?

Exocomp:
@deniszh basically to put a delay in writing so you get the effect of "batch" writes.  I like the idea of minimum data point per update but I do see how timesorted with MIN_TIMESTAMP_LAG can help - would be good if we had both (if I'm thinking about it right).

Exocomp:
@deniszh I should add that I'm working with a system where the disk is not able to keep up with the small writes, it is basically killing all its throughput.

deniszh:
@Exocomp : TBH I do not really understand how you can have both - for now I just exclude MDPU from timesorted strategy. 
How do we need to combine it - would we 
1) select metrics which have more than MIN_DATAPOINTS_PER_UPDATE datapoints in cache _AND_ already MIN_TIMESTAMP_LAG seconds
or 
2) select metrics which have more than MIN_DATAPOINTS_PER_UPDATE datapoints in cache _OR_ already MIN_TIMESTAMP_LAG seconds ?

deniszh:
Or maybe it's better to implement MPDU only in separate strategy, and not for all strategies? I'm worried a bit that MPDU trying to change the default behavior of all strategies now...

Exocomp:
@deniszh 

> select metrics which have more than MIN_DATAPOINTS_PER_UPDATE datapoints in cache OR already MIN_TIMESTAMP_LAG seconds

I like this _greedy_ method.

deniszh:
@Exocomp : OK, added that grredy logic to timesorted strategi in #692. Completely not tested, though - any feedback appreciated.

Exocomp:
Sweet!

deniszh:
Well, current implementation should be fixed - with default MIN_DATAPOINTS_PER_UPDATE == 1 it's effectively turns timesorted strategy to sorted. I'm still thinking about implementing MDPU only for selected strategies, e.g. Max and Timesorted.

PR #586, opened 2016-09-06
Set a default boolean type for AMQP_METRIC_NAME_IN_BODY

smn:
When setting `AMQP_METRIC_NAME_IN_BODY = False` in `carbon.conf` it is parsed as `'False'` (string) because it isn't recognised as a boolean type. This results in it resolving to `True` in https://github.com/graphite-project/carbon/blob/master/lib/carbon/amqp_listener.py#L118. This then raises a `ValueError` because it's expecting 3 items instead of 2 when splitting the line.


deniszh:
LGTM, merging

smn:
Wow, that was quick. Thanks!

Issue #560, opened 2016-06-02
Aggregated Metric seems to be fed back into Aggregator...


rewolf:
I have the following aggregation rule:
`abc.prod.ALL.<service>.<metric>.count (60)   = sum abc.local.*.<service>.<<metric>>.count`
Given metrics like:
`abc.prod.host1.aservice.ametric.count`
`abc.prod.host2.aservice.ametric.count`
I would expect them to be aggregated to
`abc.prod.ALL.aservice.ametric.count`

But that metric is never created.  In aggregator logs, I see
> Allocating new metric buffer for abc.prod.ALL.aservice.ametric.count

but it's not created.  If I add a layer to the generated metric like:
`abc.prod.extralayer.ALL.<service>.<metric>.count (60)   = sum abc.local.*.<service>.<<metric>>.count`
then we seem to get a recursive explosion of created metrics like:
```
abc.prod.extralayer.ALL.aservice.ametric.count
abc.prod.extralayer.ALL.ALL.aservice.ametric.count
abc.prod.extralayer.ALL.ALL.ALL.aservice.ametric.count
abc.prod.extralayer.ALL.ALL.ALL.ALL.aservice.ametric.count
```
Which led me to believe that the generated metric is then aggregated again...

I added a logging line to `AggregationProcessor.process`:
```
      else:
        log.clients("Found aggregate " + aggregate_metric + " for " + metric)
        aggregate_metrics.add(aggregate_metric)
```
And then tried with my original, desired rule.. and I eventually started to see, loglines like:
> Found aggregate abc.prod.ALL.aservice.ametric.count for abc.prod.ALL.aservice.ametric.count

It matched itself as if it was a new incoming metric... Why is it being fed back into the aggregator?

iksaif:
I think we've hit the same issue, we've fixed it with https://github.com/criteo-forks/carbon/commit/825693c97e57c7cdc36d0b0d758694010ce2be53

deniszh:
@iksaif - nice!
Should we include that fix in carbon or it has some side-effects?

obfuscurity:
It appears this is a duplicate of #455.

iksaif:
@deniszh it does have the side affect of not feedback the values in the aggregator (which some people may be depending on). My patch restore the same behavior that 0.9.x had, so feed free to merge it.

obfuscurity:
Fixed by 69f5d0e11e258da01328068f725c13335e5b8b9c.

PR #595, opened 2016-09-20
Dedicated pipeline for aggregator generated points

### LINKED ISSUE #455: carbon-aggregator not saving values
I have a simple carbon setup with the following schema:
> [anaflow]
> pattern = ^anaflow\.
> retentions = 15m:30d

and I have configured the aggregation rule

> anaflow.filesystem.filesystem.all.FlowElements.\<app_metric\> (900) = sum anaflow.filesystem.filesystem.*.FlowElements.\<app_metric\>

When sending data to the aggregator port, i can see allocations of buffers for my data in the aggregator.log and I see established and cleanly closed connections in the listener.log. The (unaggregated) data is forwarded to the cache and saved nicely, because I've set `FORWARD_ALL = True`; but I never see the aggregated data, not in graphana nor in the filesystem.

obfuscurity:
Pulling in @iksaif's fix from https://github.com/criteo-forks/carbon/commit/825693c97e57c7cdc36d0b0d758694010ce2be53 as mentioned in #560.

Fixes #455 #560 

From his commit message:
```
Generated points should not be redirected to the normal
pipeline as this creates infinite loops if, for example, the
name of the aggregated metric is the same as the name of the
original metrics. Instead setup a pipeline that just relays
generate points to the next hop.

This is effectively the behavior of carbon 0.9.*.
```

### DIFF
--- a/lib/carbon/events.py
+++ b/lib/carbon/events.py
@@ -24,8 +24,6 @@ class Event:
 
 metricReceived = Event('metricReceived')
 metricGenerated = Event('metricGenerated')
-specialMetricReceived = Event('specialMetricReceived')
-specialMetricGenerated = Event('specialMetricGenerated')
 cacheFull = Event('cacheFull')
 cacheSpaceAvailable = Event('cacheSpaceAvailable')
 pauseReceivingMetrics = Event('pauseReceivingMetrics')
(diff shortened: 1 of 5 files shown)


obfuscurity:
Works as intended in my local tests.

obfuscurity:
cc @graphite-project/committers

bmhatfield:
This seems reasonable on the surface to me, but I don't have a strong understanding of this part of the system.

deniszh:
Same, but looks reasonable. 👍

PR #596, opened 2016-09-21
Implement the Ceres storage backend against TimeSeriesDatabase

iain-buclaw-sociomantic:
With the other TimeSeriesDatabase changes now in.  Adding ceres (or any other backend) support into carbon is relatively straightforward.

This patch implements Ceres support.


obfuscurity:
LGTM

iain-buclaw-sociomantic:
I should also say that what's missing are ceres maintenance files, some of which are found in the megacarbon branch here:  https://github.com/graphite-project/carbon/tree/megacarbon/plugins/maintenance

obfuscurity:
@iain-buclaw-sociomantic Planning a separate PR for those?

iain-buclaw-sociomantic:
Can do if you think it's best to have it in carbon, maybe under some third party directory?  I don't know how integrated you'd like ceres to be. :-)

In any case, after the maintenance files, I guess that's the megacarbon branch finished.

obfuscurity:
Nah, I'd assume we want that to remain in the Ceres project, but in master.

iain-buclaw-sociomantic:
OK, so I'll check that there are no dependencies on carbon (these are standalone maintenance plugins) and raise the patch against graphite-project/ceres.

iain-buclaw-sociomantic:
Thanks.

PR #576, opened 2016-08-29
Missing cache/whisper settings checks found in 0.9.x

obfuscurity:
These were previously overlooked when their respective features were ported over from 0.9.x.

PR #598, opened 2016-09-23
Remove whisper imports from carbon (partial revert of #576)

iain-buclaw-sociomantic:
FYI @obfuscurity.  Partial revert of #576.

The whisper is config handling is found here https://github.com/graphite-project/carbon/blob/a8435b0e68ddfd8181113f18cb44c9c1adcce914/lib/carbon/database.py#L67-L96

Let's see how long until another whisper import creeps back into carbon (other than in carbon.database :-)

### DIFF
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -21,7 +21,6 @@ from os.path import join, dirname, normpath, exists, isdir
 from optparse import OptionParser
 from ConfigParser import ConfigParser
 
-import whisper
 from carbon import log, state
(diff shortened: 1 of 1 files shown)


obfuscurity:
Gah, thanks for that @iain-buclaw-sociomantic. Not sure how I missed those, I'm certain I did a `git grep` looking for those settings.

iain-buclaw-sociomantic:
@iksaif - I'm all ears on that.  So far the abstraction seems sufficient enough.

PR #599, opened 2016-10-04
instrumentation: also record 'destinations.*' stats for aggregators

iksaif:
Aggregators are also relaying points, this enable 'destinations.*'
stats for aggregators which are critical to debug queue related issues.

### DIFF
--- a/lib/carbon/instrumentation.py
+++ b/lib/carbon/instrumentation.py
@@ -129,6 +129,9 @@ def recordMetrics():
   # relay metrics
   else:
     record = relay_record
+
+  # shared relay stats for relays & aggregators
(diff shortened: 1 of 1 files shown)


lib/carbon/instrumentation.py
```diff
@@ -129,6 +129,8 @@ def recordMetrics():
   # relay metrics
   else:
     record = relay_record
+
```
obfuscurity: Would be nice to add a comment:
```
# shared relay stats for relays & aggregators
```

iksaif: done

obfuscurity:
Thanks!

PR #605, opened 2016-10-12
Add CERES_NODE_CACHING_BEHAVIOR setting.

iain-buclaw-sociomantic:
Utilizes a new configurable in Ceres, which can reduce memory consumption of
carbon-caches.  With the side effect of maybe saturating the disk, reducing
the overall throughput of persisting metrics.

As per comment, the only reason to do this is if the storage devices are fast
enough, or all in memory (in which case, you wouldn't want it cached twice).

See https://github.com/graphite-project/ceres/pull/40


obfuscurity:
LGTM

PR #614, opened 2016-12-15
Enforce better syslog tag

iksaif:


obfuscurity:
Examples of before and after?

iksaif:
after, you get something like `carbon-cache-4[4685]: [creates] new metric foo.bar`, before you just get `twisted` I think (for `carbon-cache`, but no instance and pid)

obfuscurity:
👍

PR #628, opened 2017-02-16
Fix CACHE_WRITE_STRATEGY and add TimeSortedStrategy

iksaif:
Due to a dependency hell the cache_write_strategy was always being
set to 'sorted'. Instantiate MetricCache lazilly to fix that.

Also add a new TimeSortedStrategy that will try to minize the ingestion
time.

### DIFF
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -246,7 +246,7 @@ class CarbonCacheOptions(usage.Options):
             print "Error: missing required config %s" % storage_schemas
             sys.exit(1)
 
-        if settings.CACHE_WRITE_STRATEGY not in ('sorted', 'max', 'naive'):
+        if settings.CACHE_WRITE_STRATEGY not in ('timesorted', 'sorted', 'max', 'naive'):
             log.err("%s is not a valid value for CACHE_WRITE_STRATEGY, defaulting to %s" %
                     (settings.CACHE_WRITE_STRATEGY, defaults['CACHE_WRITE_STRATEGY']))
         else:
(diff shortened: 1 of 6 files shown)


deniszh:
Fixes are looked good, but maybe you can add comments to https://github.com/graphite-project/carbon/blob/master/conf/carbon.conf.example about timesorted strategy?

iksaif:
Added some doc

deniszh:
Looks superb, @iksaif !

lib/carbon/conf.py
```diff
@@ -246,7 +246,7 @@ def postOptions(self):
             print "Error: missing required config %s" % storage_schemas
             sys.exit(1)
 
-        if settings.CACHE_WRITE_STRATEGY not in ('sorted', 'max', 'naive'):
+        if settings.CACHE_WRITE_STRATEGY not in ('timesorted', 'sorted', 'max', 'naive'):
```
DanCech: Is there a reason `random` isn't included here?  It seems like this could be avoided (and code in `cache.MetricCache()` simplified to just `write_strategy = write_strategies.get(settings.CACHE_WRITE_STRATEGY)`) if we defined `cache.write_strategies` as a `dict` mapping the string values to classes and instead checked if `settings.CACHE_WRITE_STRATEGY` was a key in that `dict`.

iksaif: Totally true. I'll clean that up in another PR :) thanks for noticing !

iain-buclaw-sociomantic: From what I recall, random strategy has linear complexity, and so makes it terrible for high volume workloads.  Same for max strategy too.

https://github.com/graphite-project/carbon/blob/abff76a3962515003961f40f9e5fd8a3a28ed5c6/lib/carbon/tests/benchmark_cache.py#L73-L76

PR #629, opened 2017-02-16
writer: sleep less

iksaif:
a lot of things can happen in one second, and
it is also our minimum resolution.


DanCech:
Seems reasonable 👍

deniszh:
:+1:

PR #601, opened 2016-10-05
Make protocols pluggable

iksaif:
- A server protocol just need to implement a build(root_service) class method that will add its service to the root service.
- A client protocol will need to inherit from CarbonClientFactory and implement clientProtocol().

This code was *not* tested with amqp and manhole because there are no unit tests and I do not have the infrastructure to test these.

First step towards #275.

PR #635, opened 2017-03-26
Fix carbon line client protocol so it can send correct data

waltzofpearls:
Currently when `DESTINATION_PROTOCOL` is set to `line` for carbon-relay.
It sends incorrect data to the destination. The order of metric value
and timestamp are changed from the source, for example, it turns
"test.mytest 1 1490397060" into "test.mytest 1490397060 1".

`CarbonLineClientProtocol` was introduced in a previous pull request #601
to support pluggable client protocols along with a number of other
changes. The incorrect order takes place in its method
`_sendDatapointsNow()`.

This fix will correct the order of segments, so metrics path, metric
value and timestamp will be placed in the right order. An unit test is
also added to ensure carbon line client protocol will send correct data
in the future.

### DIFF
--- a/lib/carbon/client.py
+++ b/lib/carbon/client.py
@@ -358,7 +358,7 @@ class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
 
   def _sendDatapointsNow(self, datapoints):
     for metric, datapoint in datapoints:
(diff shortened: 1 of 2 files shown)


deniszh:
Ouch. @iksaif ?

PR #639, opened 2017-04-03
aggregator: allow to use a bounded expiring cache to limit memory usage

iksaif:
Depend on https://github.com/graphite-project/carbon/pull/638

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -559,3 +559,12 @@ MAX_AGGREGATION_INTERVALS = 5
 # Specify the user to drop privileges to
 # If this is blank carbon-aggregator runs as the user that invokes it
 # USER =
+
+# Part of the code, and particularly aggregator rules, need
+# to cache metric names. To avoid leaking too much memory you
+# can tweak the size of this cache. The default allow for 1M
+# different metrics per rule (~200MiB).
+# CACHE_METRIC_NAMES_MAX=1000000
+
+# You can optionally set a ttl to this cache.
+# CACHE_METRIC_NAMES_TTL=600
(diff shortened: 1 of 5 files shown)


wdauchy:
I guess there is one commit which should not be here (	Use requirements.txt for tox) since it is already present in the previous MR
:-1:

iksaif:
It's because it depends on the other one. I don't know how to do 'dependent' pull requests on github. Once the other is merged the requirements.txt commit will disappear (except the part with cachetools)

deniszh:
@iksaif: I also don't know how to properly did dependent PRs.
https://stackoverflow.com/questions/26619478/are-dependent-pull-requests-in-github-possible

Adding changes from other commits doesn't help either, this PR should be rebased anyway.

deniszh:
Adding `cachetools` as another mandatory module? Meh.
Could we use something lightweight but "built-in in carbon code"?

iksaif:
IMOH cachetools is very lightweight, it contains only what is necessary and is also compatible with all versions of python and all OSes.

Re-implementing an expiring LRU cache is tricky, and I tried write something for carbon only, but I ended up duplicating most of what cachetools is doing :/

iksaif:
FYI: I tried this patch on a pretty load environment and it seems to cap the used memory as expected:

before:
carbon-before

after:
carbon-after

iksaif:
Done.
I'd keep that in master :).

wdauchy:
:+1:

deniszh:
Ah, `carbon.conf` ofc. 🤦‍♂️

iksaif:
Any objection ?

deniszh:
No objection

PR #640, opened 2017-04-05
protocols: enable traffic shaping with ShapedProtocolFactory()

iksaif:
This is using https://github.com/twisted/twisted/blob/HEAD/src/twisted/protocols/htb.py
to try to apply some traffic shaping to incoming connections.


deniszh:
Traffic shaping? Interesting.
Does it really helps with some issues, @iksaif ?

iksaif:
I'll add something to carbon.conf.example and try to get a graph that shows how it helps.

The idea is that we had some very big users sending bursts of ~300MiB of metrics and this would be starving others when these metrics had to be relayed on inter-continental links on the next hop.

This shapes the burst and make the metric receiver more fair.

deniszh:
Cool, looks nice!

deniszh:
Still want to improve that, @iksaif ?

iksaif:
Getting some graphs is still on my todo list, I'll ping this issue when ready :)

iksaif:
Probably will never get this done

PR #641, opened 2017-04-05
Add a benchmark for the aggregator

iksaif:


iksaif:
Fixed some codacy issues. The others will need more refactoring and I'll do them in another commit.

PR #647, opened 2017-04-24
graphite aggregating sum wrongly with example conf

pschuermann:
See the example here: https://answers.launchpad.net/graphite/+question/286462


deniszh:
Yeah, completely forgot about that issue. Thanks!

PR #649, opened 2017-04-27
protocols: support SO_REUSEPORT

wdauchy:
this will avoid the need to bind each carbon process to a different port
and balance the traffic with iptables rules

this requires linux >= 3.9

patches are on their way to support reuseport directly in twisted

### DIFF
--- a/lib/carbon/protocols.py
+++ b/lib/carbon/protocols.py
@@ -1,8 +1,11 @@
 import time
+import socket
 
 from twisted.internet.protocol import ServerFactory, DatagramProtocol
 from twisted.application.internet import TCPServer, UDPServer
+from twisted.application import service
 from twisted.internet.error import ConnectionDone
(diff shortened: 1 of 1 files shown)


iksaif:
That would be super useful ! @dzhdanov what do you think ?
@wdauchy  maybe add some comments with a link to the twisted PR in the code ?

wdauchy:
@iksaif comment added in class definition

iksaif:
(please wait for a review from somebody else :) )

deniszh:
It's backward compatible, right?

wdauchy:
@deniszh yes it is; even if users put same port on different relay config, it will just work if they use linux >= 3.9

wdauchy:
@deniszh FYI I had to slightly adapt my commit for 1.0.x, see https://github.com/criteo-forks/carbon/commit/fbe31940c7f26eef852c80dde04873cc6da30d48

PR #653, opened 2017-05-08
Making queue consuming logs configurable

deniszh:
Trying to fix https://answers.launchpad.net/graphite/+question/631657

Not sure if LOG_CACHE_QUEUE_SORTS is good variable for it, though.


iksaif:
I have that doing more or less the same thing: https://github.com/criteo-forks/carbon/commit/84d0be8519f50fdbf38cebe1744e14f230f3f3df

Feel free to steal some of the additional bits :)

deniszh:
Yep, your patch is better, stealing it :)

deniszh:
Will cherry pick to 1.0.x, just in case

PR #656, opened 2017-05-15
Initial Python 3 compatibility

piotr1212:
Adds initial python3 compatibility. All unittests and benchmark now succeed on py27, py34, py35, py36. But needs real-life testing.

While updating the GRAPHITE_NO_PREFIX part in setup.py I noticed that the original didn't work on my system with Python 2.7. It would just write an empty setup.cfg file. Instead of fixing the original code I just removed that part as I understand from comments in other issues that it is not needed anymore and I personally think the prefix should not be the default behaviour.

AMQP is disabled on python3 as the modules are not compatible yet.

mocker is also not compatible yet, I've added a simplified makeFile function to the testcase and removed the import of mocker.

For the hashring the "node:instance" string needs to be encoded, I've encoded it with ascii as I was not sure if UTF-8 would cause trouble somewhere else.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -23,7 +23,8 @@ from carbon import events, log
 from carbon.pipeline import Processor
 
 
-def by_timestamp((timestamp, value)):  # useful sort key function
+def by_timestamp(t_v): # useful sort key function
+  (timestamp, _) = t_v
   return timestamp
 
 
@@ -57,14 +58,14 @@ class NaiveStrategy(DrainStrategy):
 
     def _generate_queue():
       while True:
-        metric_names = self.cache.keys()
+        metric_names = list(self.cache.keys())
         while metric_names:
           yield metric_names.pop()
 
     self.queue = _generate_queue()
 
   def choose_item(self):
-    return self.queue.next()
+    return next(self.queue)
 
 
 class MaxStrategy(DrainStrategy):
@@ -80,7 +81,7 @@ class MaxStrategy(DrainStrategy):
 class RandomStrategy(DrainStrategy):
   """Pop points randomly"""
   def choose_item(self):
-    return choice(self.cache.keys())
+    return choice(list(self.cache.keys()))
 
 
 class SortedStrategy(DrainStrategy):
@@ -105,7 +106,7 @@ class SortedStrategy(DrainStrategy):
     self.queue = _generate_queue()
 
   def choose_item(self):
-    return self.queue.next()
+    return next(self.queue)
 
 
 class TimeSortedStrategy(DrainStrategy):
@@ -130,7 +131,7 @@ class TimeSortedStrategy(DrainStrategy):
     self.queue = _generate_queue()
 
   def choose_item(self):
-    return self.queue.next()
+    return next(self.queue)
 
 
 class _MetricCache(defaultdict):
@@ -174,7 +175,7 @@ class _MetricCache(defaultdict):
       metric = self.strategy.choose_item()
     else:
       # Avoid .keys() as it dumps the whole list
-      metric = self.iterkeys().next()
+      metric = next(iter(self.keys()))
     return (metric, self.pop(metric))
 
   def get_datapoints(self, metric):
--- a/lib/carbon/protocols.py
+++ b/lib/carbon/protocols.py
@@ -134,15 +134,13 @@ class MetricReceiver(CarbonServerProtocol, TimeoutMixin):
       events.resumeReceivingMetrics.removeHandler(self.resumeReceiving)
 
   def metricReceived(self, metric, datapoint):
-    if settings.USE_WHITELIST:
-      if metric in BlackList:
-        instrumentation.increment('blacklistMatches')
-        log.debug("metric {} dropped: in blacklist")
-        return
-      if metric not in WhiteList:
-        instrumentation.increment('whitelistRejects')
-        log.debug("metric {} dropped: not in whitelist")
-        return
+    if BlackList and metric in BlackList:
+      instrumentation.increment('blacklistMatches')
+      return
+    if WhiteList and metric not in WhiteList:
+      instrumentation.increment('whitelistRejects')
+      return
+
     if datapoint[1] != datapoint[1]:  # filter out NaN values
       return
     if int(datapoint[0]) == -1:  # use current time if none given: https://github.com/graphite-project/carbon/issues/54
--- a/lib/carbon/protocols.py
+++ b/lib/carbon/protocols.py
@@ -13,6 +13,7 @@ from carbon.conf import settings
 from carbon.regexlist import WhiteList, BlackList
 from carbon.util import pickle, get_unpickler
 from carbon.util import PluginRegistrar
+from six import with_metaclass
 
 
 class CarbonReceiverFactory(ServerFactory):
@@ -63,8 +64,7 @@ class CarbonService(service.Service):
     def stopService(self):
         self._port.stopListening()
 
-class CarbonServerProtocol(object):
-  __metaclass__ = PluginRegistrar
+class CarbonServerProtocol(with_metaclass(PluginRegistrar, object)):
   plugins = {}
 
   @classmethod
@@ -133,12 +133,15 @@ class MetricReceiver(CarbonServerProtocol, TimeoutMixin):
       events.resumeReceivingMetrics.removeHandler(self.resumeReceiving)
 
   def metricReceived(self, metric, datapoint):
-    if BlackList and metric in BlackList:
-      instrumentation.increment('blacklistMatches')
-      return
-    if WhiteList and metric not in WhiteList:
-      instrumentation.increment('whitelistRejects')
-      return
+    if settings.USE_WHITELIST:
+      if metric in BlackList:
+        instrumentation.increment('blacklistMatches')
+        log.debug("metric {} dropped: in blacklist")
+        return
+      if metric not in WhiteList:
+        instrumentation.increment('whitelistRejects')
+        log.debug("metric {} dropped: not in whitelist")
+        return
     if datapoint[1] != datapoint[1]:  # filter out NaN values
       return
     if int(datapoint[0]) == -1:  # use current time if none given: https://github.com/graphite-project/carbon/issues/54
@@ -177,7 +180,8 @@ class MetricDatagramReceiver(MetricReceiver, DatagramProtocol):
 
     super(MetricDatagramReceiver, cls).build(root_service)
 
-  def datagramReceived(self, data, (host, port)):
+  def datagramReceived(self, data, addr):
+    (host, _) = addr
     for line in data.splitlines():
       try:
         metric, value, timestamp = line.strip().split()
@@ -208,7 +212,7 @@ class MetricPickleReceiver(MetricReceiver, Int32StringReceiver):
     for raw in datapoints:
       try:
         (metric, (value, timestamp)) = raw
-      except Exception, e:
+      except Exception as e:
         log.listener('Error decoding pickle: %s' % e)
 
       try:
@@ -239,7 +243,7 @@ class CacheManagementHandler(Int32StringReceiver):
     cache = MetricCache()
     if request['type'] == 'cache-query':
       metric = request['metric']
-      datapoints = cache.get(metric, {}).items()
+      datapoints = list(cache.get(metric, {}).items())
       result = dict(datapoints=datapoints)
       if settings.LOG_CACHE_HITS:
         log.query('[%s] cache query for \"%s\" returned %d values' % (self.peerAddr, metric, len(datapoints)))
@@ -249,7 +253,7 @@ class CacheManagementHandler(Int32StringReceiver):
       datapointsByMetric = {}
       metrics = request['metrics']
       for metric in metrics:
-        datapointsByMetric[metric] = cache.get(metric, {}).items()
+        datapointsByMetric[metric] = list(cache.get(metric, {}).items())
 
       result = dict(datapointsByMetric=datapointsByMetric)
 
--- a/lib/carbon/protocols.py
+++ b/lib/carbon/protocols.py
@@ -65,6 +65,7 @@ class CarbonService(service.Service):
     def stopService(self):
         self._port.stopListening()
 
+
 class CarbonServerProtocol(with_metaclass(PluginRegistrar, object)):
   plugins = {}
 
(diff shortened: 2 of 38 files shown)


piotr1212:
Hmm, sorry, found some issues with this PR. I'll do some testing and report when I think there are no issues left.

piotr1212:
I've now got most use cases working/tested, that is:

cache
relay
aggregator
line input
udp input
pickle input
query interface
pickle output

I've hardcoded pickle output protocol version to 2 as this makes interoperability with graphite-web, carbons running on py2 and pypy possible. I could make this configurable but don't really see the need for that.

I've assumed that Graphite line protocol only should support ascii characters. I don't know if there is need for UTF-8 or if there is some formal graphite protocol specification. carbon-c-relay only allows ascii.

deniszh:
Could you please fix codacity issues - https://www.codacy.com/app/graphite-project/carbon/pullRequest?prid=650788 ? 
At least, unused variables. (replace `value` with `_`)

setup.py
```diff
@@ -65,34 +29,31 @@
...
+    data_files=install_files,
+    install_requires=['Twisted', 'txAMQP', 'cachetools'],
+    classifiers=(
         'Intended Audience :: Developers',
         'Natural Language :: English',
         'License :: OSI Approved :: Apache Software License',
         'Programming Language :: Python',
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 2 :: Only',
```
deniszh: should it be "Python :: 2 :: Only" if python 3 is supported?

lib/carbon/cache.py
```diff
@@ -23,7 +23,8 @@
 from carbon.pipeline import Processor
 
 
-def by_timestamp((timestamp, value)):  # useful sort key function
+def by_timestamp(t_v): # useful sort key function
+  (timestamp, value) = t_v
```
deniszh: timestamp, _ = t_v

lib/carbon/protocols.py
```diff
@@ -177,13 +180,14 @@ def build(cls, root_service):
 
     super(MetricDatagramReceiver, cls).build(root_service)
 
-  def datagramReceived(self, data, (host, port)):
+  def datagramReceived(self, data, addr):
+    (host, port) = addr
```
deniszh: host, _ = addr ?

deniszh:
>I've hardcoded pickle output protocol version to 2 as this makes interoperability with graphite-web and carbons running on py2 and pypy possible. I could make this configurable but don't really see the need for that.

Also not sure, let's stick to 2.

>I've assumed that Graphite line protocol only should support ascii characters. I don't know if there is need for UTF-8 or if there is some formal graphite protocol specification. carbon-c-relay only allows ascii.

Also not sure about protocol spec. Maybe somebody else knows.
/cc @obfuscurity @iksaif @DanCech ?

DanCech:
For interoperability's sake ASCII is probably the way to go, character set issues are a real pain and UTF-8 support can be very patchy.

iksaif:
looks good

drawks:
FWIW, if an earlier version (in the same major release) accepted metric keys that were encoded other than ascii and this new version limits to ascii, that is a decent case for bumping the major version. I'm sure that breaking basic functionality for people using keys that are meaningfully named using non-ascii chars would be bad times.

Let's make sure we're being good maintainers that don't eliminate, for instance, international users for the sake of simplicity without at a minimum communicating the intended change and waiting for feedback on the impact of that change.

deniszh:
I agreed with @drawks - if it changes behavior it should be or configurable or next major release.

piotr1212:
Just did a quick check on master (py2): Carbon seems to work fine with UTF-8, but graphite-web throws an exception. I'll do some more testing later today.

iksaif:
The grammar of graphite is supposed to be defined by https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/grammar.py

```
printables = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\\
'()*+,-./:;<=>?@[\\]^_`{|}~'
validMetricChars = ''.join((set(printables) - set(symbols)))
```

So I think only ASCII is explicitely supported.

drawks:
That only covers graphite-web, if carbon accepts it and whisper writes it, then it doesn't matter if graphite-web likes it or not, because a not inconsequential number of user don't use graphite-web, but do use carbon and whisper.

deniszh:
I'm wondering is it possible to stick with python2-way of string processing somehow?
@piotr1212

piotr1212:
> I'm wondering is it possible to stick with python2-way of string processing somehow?
> @piotr1212

I thought Python 2 strings were ascii and that carbon couldn't handle UTF-8. But obviously I was wrong. I think if I just switch the encoding to UTF-8 we would get the old behaviour. I was worried that UTF-8 might mix up the hashring calculation of metrics, but that does not make sense to me now.

deniszh:
I made a quick test, @piotr1212 

python2:
```
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from md5 import md5
>>> a = "blah"
>>> print str(a)
blah
>>> print md5(str(a)).hexdigest()
6f1ed002ab5595859014ebf0951522d9
>>> a = "жопа"
>>> print str(a)
жопа
>>> print md5(str(a)).hexdigest()
ec3459df2701beb8334357d9ba7f283d
>>> print a.encode('utf-8')
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
>>> print md5(a.encode('utf-8')).hexdigest()
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
>>>
```

python3:
```
Python 3.6.1 (default, Mar 23 2017, 16:49:06)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from hashlib import md5
>>> a = "blah"
>>> print(str(a))
blah
>>> print(md5(str(a)).hexdigest())
  File "<stdin>", line 1, in <module>
TypeError: Unicode-objects must be encoded before hashing
>>> print(a.encode('utf-8'))
b'blah'
>>> print(md5(a.encode('utf-8')).hexdigest())
6f1ed002ab5595859014ebf0951522d9
>>> a = "жопа"
>>> print(a.encode('utf-8'))
b'\xd0\xb6\xd0\xbe\xd0\xbf\xd0\xb0'
>>> print(md5(a.encode('utf-8')).hexdigest())
ec3459df2701beb8334357d9ba7f283d
>>>
```

Good news - hash are the same on ASCII on python2 and 3
Bad news - `a.encode('utf-8')` is not working on non-ascii strings in python2

piotr1212:
That seems to be the difference in str and unicode objects:
```
Python 2.7.13 (default, May 10 2017, 20:04:28) 
[GCC 6.3.1 20161221 (Red Hat 6.3.1-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> a = "жопа"
>>> print(a.encode('utf-8')
... )
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
>>> 
>>> 
>>> a = "жопа"
>>> print(a.encode('utf-8'))
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 0: ordinal not in range(128)
>>> 
>>> b = u"жопа"
>>> print(b.encode('utf-8'))
жопа
>>> 
>>> type(a)
<type 'str'>
>>> 
>>> type(b)
<type 'unicode'>
```

I'll add some utf unittest cases.

deniszh:
Indeed. But are you sure that key are coming in u'...' form if people are using utf-8 in metric names? I doubt so. I think they're coming as normal bytestreams. str() happily ignores that, but encode() is not.

piotr1212:
> But are you sure that key are coming in u'...' form if people are using utf-8 in metric names? I doubt so. I think they're coming as normal bytestreams. str() happily ignores that, but encode() is not.

For line and udp input the conversion from bytes to str objects with utf-8 encoding happens earlier in the pipeline (protocols.py). It doesn't matter if the bytes are ascii or utf-8 as a utf-8 byte is larger than 128 and ascii lower. No clue what happens with latin1 encoding though...

But good point: I've checked pickle input and this seems broken now (for relay at least, cache works fine, haven't checked the aggregator yet).

piotr1212:
benchmark_routers.py shows a 5 to 10 % performance decrease when I switch from str to unicode objects in Python 2.7. The benchmark does not cover the decoding of the input in protocols.py so real performance impact will probably be even higher. I'll have to find some other way to do this.

piotr1212:
> Bad news - a.encode('utf-8') is not working on non-ascii strings in python2

The issue here seemed that the string was already encoded and python then first decodes is before encoding resulting in something like: ```a.decode('ascii').encode('utf-8')``` which fails. I couldn't find a portable version for this so just conditionalized it.

I've done a lot of testing, this seems the way to go. It shouldn't change the behaviour on py2 and use native 'str' in py3. I tried if it would be faster to switch to bytearrays in py3 but this messed up the code and was actually slower..

Should I rebase+squash some commits? Commit history is a mess now, I promise to do some more testing next time before creating a PR.

deniszh:
Hi @piotr1212,
Looks great, thanks a lot for your contribution! Python 3 is a way to go anyway, we should be there sooner or later.
If you have a good plan how to squash commits properly - please do. Otherwise  I'll make "squash and merge".

DanCech:
For what it's worth the grammar parser in graphite-web expects series names to be a subset of ascii, so series with non-ascii characters are unlikely to be useful.On Jun 6, 2017 10:10 AM, Denis Zhdanov <notifications@github.com> wrote:Hi @piotr1212,
Looks great, thanks a lot for your contribution! Python 3 is a way to go anyway, we should be there sooner or later.
If you have a good plan how to squash commits properly - please do. Otherwise  I'll make "squash and merge".

piotr1212:
@DanCech Any clue why that restriction is put in place? I've extensively tested carbon on utf-8 and found no issues. Did a quick hack of graphite-web to allow utf-8 and besides a minor issue with Chinese characters not showing up in the legend it works well. 

https://github.com/piotr1212/graphite-web/commit/17e23efecb6cb40059264fabd3ea36718065176e 
https://imgur.com/LXpM1Si

Chinese don't show up because the selected font does not have the glyphs. To support it properly we would need to switch to pango which can fallback to a different font if a specific glyph is not present.

drawks:
> For what it's worth the grammar parser in graphite-web expects series names to be a subset of ascii, so series with non-ascii characters are unlikely to be useful.

As I mentioned before, graphite-web and carbon are two decoupled pieces. If non-ascii metric names work in carbon we shouldn't break that feature with the justification that they don't work in graphite-web. There is definitely a subset of user which use carbon without using graphite-web.

If at all feasible the best solution would be fix graphite-web to support non-ascii metric names. Increase I18N not throw it away as a bug.

lib/carbon/protocols.py
```diff
@@ -208,14 +219,18 @@ def stringReceived(self, data):
...
+      except Exception as e:
         log.listener('Error decoding pickle: %s' % e)
 
       try:
         datapoint = (float(value), float(timestamp))  # force proper types
       except ValueError:
         continue
 
+      # convert python2 unicode objects to str/bytes
+      if type(metric) is 'unicode':
```
deniszh: codacy recommends to replace with
`if isinstance(metrics, 'unicode'):`

deniszh:
👍 what @drawks says - we should keep compatibility, because `carbon` is a separate project.
But as far as I understand now it's compatible with old behavior, despite python version. Am I right, @piotr1212 ?

DanCech:
Yes, what I'm saying is not that utf8 support is a bad thing (I'm a huge proponent of it) but that if it doesn't work today it shouldn't be a blocker for this pr.On Jun 6, 2017 2:26 PM, Dave Rawks <notifications@github.com> wrote:
For what it's worth the grammar parser in graphite-web expects series names to be a subset of ascii, so series with non-ascii characters are unlikely to be useful.

As I mentioned before, graphite-web and carbon are two decoupled pieces. If non-ascii metric names work in carbon we shouldn't break that feature with the justification that they don't work in graphite-web. There is definitely a subset of user which use carbon without using graphite-web.
If at all feasible the best solution would be fix graphite-web to support non-ascii metric names. Increase I18N not throw it away as a bug.

piotr1212:
> 👍 what @drawks says - we should keep compatibility, because carbon is a separate project.
But as far as I understand now it's compatible with old behavior, despite python version. Am I right, @piotr1212 ?

Well, Yes.... but not 100% accurate. Both implementations support UTF-8 / ASCII. The corner cases which don't work when running in py3 are other encodings, for example latin1.

The PR keeps the Python2 implementation as is. The py2 imp just treats the input as byte arrays and doesn't care about encoding at all. The Python3 implementation assumes input is in UTF-8, which is backwards compatible with ASCII, but not with (for example) latin1. I did an attempt to get the py3 implementation work on byte arrays was well, but this made the code a real mess and (the relay) was 4 times slower. I don't think keeping the py3 implementation compatible with latin1 is worth the trouble, if people actually want to use latin1, they will most likely not be running Python 3 anyway.

> If at all feasible the best solution would be fix graphite-web to support non-ascii metric names. Increase I18N not throw it away as a bug.

That is exactly what I will try when I get around to it.

lib/carbon/protocols.py
piotr1212: Hmm, this code is actually incorrect, unicode must not be quoted.

deniszh:
Any objections about merging this?
I think it's fully compatible when using python 2.x now, and we'll care about 3.x compatibility when/if converting graphite-web to 3.x
/cc @drawks @DanCech ?

iksaif:
I'm find, and it would be great to have Graphite converted to 3.0 for the next release :)

piotr1212:
last commit fixes invalid line logging in py3, adds some unittests, catches more pickle.loads exceptions.

> I think it's fully compatible when using python 2.x now, and we'll care about 3.x compatibility when/if converting graphite-web to 3.x

Although I'm quite sure I've missed something, all "happy path" scenarios I could think of seem to work on py3. (tested: cache, relay, aggregator, line input, pickle input, udp, protobuf)

iksaif:
:+1:

deniszh:
OK, let's merge this!
Thanks again, @piotr1212, for pushing that forward.

PR #659, opened 2017-05-23
Add carbon-cache-aggregator

iksaif:
This avoids network and CPU overhead between cache and aggregators since
they often receive the same metrics. Most of the CPU was found to be used in
the serializing and deserializing functions.

### DIFF
--- a/bin/carbon-aggregator-cache.py
+++ b/bin/carbon-aggregator-cache.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+"""Copyright 2009 Chris Davis
+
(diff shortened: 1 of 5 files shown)


deniszh:
Looks good, but some docs is needed.

iksaif:
I'm not sure where I could put the documentation for this. Ideas ?

deniszh:
You can put it to https://graphite.readthedocs.io/en/latest/carbon-daemons.html
IMO should me enough.

iksaif:
Created https://github.com/graphite-project/graphite-web/pull/1947

PR #661, opened 2017-06-01
Add DYNAMIC_ROUTER settings

Thib17:
cc @iksaif


drawks:
Interesting features, but I think it would be more appropriate to split them into multiple PRs. I don't see any intrinsic overlap of function or code that should make these a single PR.

Thib17:
Ok, so let's start with DYNAMIC_ROUTER settings first.

lib/carbon/client.py
```diff
@@ -306,20 +315,69 @@ def sendHighPriorityDatapoint(self, metric, datapoint):
...
+      log.clients("Adding client %s:%d:%s to router" % destination)
+      self.router.addDestination(destination)
+      state.events.resumeReceivingMetrics()
+
+  def destinationDown(self, destination):
+    # Only blacklist the destination if we tried a lot.
+    log.clients("Destination is down: %s:%d:%s (%d/%d)" % (
+        destination[0], destination[1], destination[2], self.retries,
+        settings.DYNAMIC_ROUTER_MAX_RETRIES))
+    # Retries com from the ReconnectingClientFactory.
```
deniszh: com?

deniszh:
Interesting feature, indeed! Thanks, @Thib17 !
More reviews needed. /cc @drawks @iksaif @DanCech @graphite-project/committers

conf/carbon.conf.example
```diff
@@ -370,6 +370,17 @@ DESTINATIONS = 127.0.0.1:2004
 # extended with CarbonClientFactory plugins and defaults to "pickle".
 # DESTINATION_PROTOCOL = pickle
 
+# When using consistent hashing it sometime make sense to make
```
iksaif: makes*

lib/carbon/client.py
iksaif: comes*

conf/carbon.conf.example
Thib17: Done

lib/carbon/client.py
Thib17: Done

PR #666, opened 2017-06-12
Add MIN_TIMESTAMP_LAG

Thib17:
cc @iksaif


deniszh:
Any objections about merging this?

conf/carbon.conf.example
```diff
@@ -85,6 +85,11 @@ MAX_CREATES_PER_MINUTE = 50
 # second.
 MIN_TIMESTAMP_RESOLUTION = 1
 
+# Set the minimum lag for a point to be written to the database in order to
+# optimize batching. Setting this to 0 disable the feature. This currently
+# only works when using the timesorted write strategy.
+# MIN_TIMESTAMP_LAG = 30
```
DanCech: I'd like to see this updated to match the default config, and explain exactly what enabling it does.

Thib17: Done

Issue #667, opened 2017-06-21
ImportError: No module named carbon.util


mizhdi:
When I trying to run in a virtualenv：
carbon-cache.py start

how to fix?

deniszh:
Something is wrong with your installation.
How do you installing it?
Try e.g. this:
```
virtualenv /opt/graphite
source /opt/graphite/bin/activate
cd ~ && git clone  https://github.com/graphite-project/carbon.git
cd carbon && pip install -r requirements.txt
python setup.py install --prefix=/opt/graphite --install-lib=/opt/graphite/lib
cd /opt/graphite/
cp conf/carbon.conf.example conf/carbon.conf
cp conf/storage-schemas.conf.example conf/storage-schemas.conf
./bin/carbon-cache.py start
```

mizhdi:
It works. Thank you @ deniszh

ybizeul:
Same problem here, except I'm not using virtualenv.
It seems to be related to the recent changes in https://github.com/graphite-project/carbon/commit/7516cf148c5d75c2cd066c32ad95f5d617cca2f0#diff-2eeaed663bd0d25b7e608891384b7298

Trying to install with both `--prefix=/opt/graphite --install-lib=/opt/graphite/lib` solved the problem

deniszh:
TBH I also suspected that change, but have no idea how it can be related.... Any ideas, @piotr1212 ?

deniszh:
Ah, @ybizeul is right, it's because of https://github.com/graphite-project/carbon/commit/7516cf148c5d75c2cd066c32ad95f5d617cca2f0#commitcomment-22693426
Looks like we should restore this logic.

deniszh:
Or it has clashed with python3, @piotr1212 ?

piotr1212:
Yes I removed it. The code wasn't py3 compatible, seemed like a big hack,
broke installing in other prefix than opt. Overriding it with the env
variable was not intuitive. Many new users had trouble finding out to set
the correct Python path. Opt isn't the default for python modules.
I didn't realize this would break users installations, thought it was some
old strange leftover. Will try to restore it this evening.

Op do 22 jun. 2017 09:46 schreef Denis Zhdanov <notifications@github.com>:

piotr1212:
https://github.com/graphite-project/carbon/issues/668 restores the original behaviour with installing to /opt/graphite by default.

@mizhdi @ybizeul : Were you running clean installation or upgrading older ones?

[...]

PR #668, opened 2017-06-23
Restore default "/opt/graphite" installation prefix.

piotr1212:
Restores the default installation prefix to "/opt/graphite" and libdir to
"/opt/graphite/lib". Now compatible with py2 and py3.

Instructions are now being printed to inform the user how to use system
default installation locations.

TBH: I don't really understand why the overwriting of config files is needed.

This should fix:https://github.com/graphite-project/carbon/issues/667

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,49 @@
 from __future__ import with_statement
 
 import os
-import platform
 from glob import glob
+from six.moves import configparser
+
+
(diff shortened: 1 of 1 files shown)


DanCech:
Seems like it'll do the job

deniszh:
TBH I dunno why the overwriting of config files is needed either. But as Dan said - it's do the job. Merging.

PR #669, opened 2017-06-29
fix 'retentions' key error when loading storage schema

dekdaas:
Previously throwing a key error. value incorrectly accessed, modified to match method used for parsing other options.


deniszh:
Hi @dekdaas ,
Could you please elaborate your change a bit? 'retentions' parameter is madatory, thats's why it have `try .. except` block and throwing an error. I do not see point putting `get()` there.

deniszh:
Sorry, closing it then.

PR #670, opened 2017-06-29
py3: fix iterkeys() and exception.message

piotr1212:
d.iterkeys() should be iter(d) instead of iter(d.keys()) so it doesn't return
a whole list but only the iterators.

exception.message doesn't exits in Python3


Akuli:
Nice :)

PR #672, opened 2017-07-08
Backporting fnv1a_hash ring from graphite-web to carbon 1.0.x

deniszh:
Backporting [fnv1a_hash ring from graphite-web](https://github.com/graphite-project/graphite-web/pull/1723) to  carbon 1.0.x 

It's not a full support of fnv1a hashing in carbon, only hashing part for now. I want to use that in carbonate for a rebalancing of fnv clusters - https://github.com/graphite-project/carbonate/pull/83

### DIFF
--- a/lib/carbon/hashing.py
+++ b/lib/carbon/hashing.py
@@ -4,24 +4,50 @@ except ImportError:
   from md5 import md5
 import bisect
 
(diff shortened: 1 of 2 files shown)


iksaif:
LGTM

PR #674, opened 2017-07-20
Tagging Support

DanCech:
This PR defines a mechanism for handling tagged series in graphite, and is designed to be used in conjunction with [PR 2002 in graphite-web](https://github.com/graphite-project/graphite-web/pull/2002).

It defines a format for tagged series in carbon, which is `<metric>;<tag1>=<value1>;<tag2>=<value2>`

Each logical metric in the system is uniquely identified by the combination of its metric name and set of tags.  It does not attempt to store metadata about a metric that might change over time, if a tag is changed that creates a new series in the same way as changing a node in a regular series.

This format was chosen to remain compatible with existing carbon-based tooling as much as possible, while maintaining a clear format for defining tags and avoiding any issues with tag names or values containing `.`, which can be a problem for implementations that embed tag/value pairs as nodes within the dotted series name.

The major changes to the carbon pipeline are:

- use of TaggedSeries class (shared with graphite-web, could possibly be put into a separate library) to canonicalize tagged series names by ordering tags alphabetically by tag name immediately before putting the points into the write cache
- make HTTP POST to the new `/tags/tagSeries` endpoint in graphite-web to add series to the tag database, this currently happens on every create operation and on a specified interval (`settings.TAG_UPDATE_INTERVAL` by default 100) for update operations
- refactor the writer twisted service to remove the calls to `time.sleep()`, since they block the entire thread and cause issues with the tagging requests

The biggest architectural change is that carbon now needs to be aware of the location of graphite-web (hence the new `settings.GRAPHITE_URL` configuration item) so that it can make the HTTP calls to add series to the tag database.

When tagged series are detected, they are stored in a separate hierarchy of folders under a top-level `_tagged` folder, where subfolders are created by using the first 3 hex digits of the sha256 hash of the tagged metric path (4096 possible folders), and second-level subfolders are based on the following 3 hex digits (another 4096 possible folders) for a total of 4096^2 possible subfolders.  The metric files themselves are created with any `.` in the metric path replaced with `-`, to avoid any issues where metrics, tags or values containing a '.' would end up creating further subfolders.  This is implemented the same way for both whisper and ceres, but by design is handled within the database class (and in graphite-web within the finder) so that different backends can create their own schemes if desired.

A concrete example:

`some.metric;tag1=value2;tag2=value.2`

with sha256 hash `effaae5b55950b150a7493d2f5430772ca2460bf7cccb4d9e523643df54c010f` would be stored in:

`_tagged/eff/aae/some-metric;tag1=value2;tag2=value-2.wsp` (whisper)
`_tagged/eff/aae/some-metric;tag1=value2;tag2=value-2` (ceres)


lib/carbon/writer.py
```diff
@@ -42,121 +54,188 @@
...
       try:
-        t1 = time.time()
-        # If we have duplicated points, always pick the last. update_many()
-        # has no guaranted behavior for that, and in fact the current implementation
-        # will keep the first point in the list.
-        datapoints = dict(datapoints).items()
-        state.database.write(metric, datapoints)
-        updateTime = time.time() - t1
+        state.database.create(metric, archiveConfig, xFilesFactor, aggregationMethod)
+        tagMetric(metric)
```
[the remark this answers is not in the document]
iksaif: Yes, the same way CarbonLink protocol is handled in all the finders. I know that in my implementation I will not want to call Graphite each time I create a metric because it will be too slow.

What can be done is that the default implementation of database.tagMetric() calls Graphite, but can be overloaded to do something else.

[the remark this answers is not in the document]
DanCech: The problem with doing that is that right now the call `state.database.create` (and `state.database.write`) is blocking, whereas `tagMetric` is designed to be async and use the twisted `HTTPConnectionPool` to avoid blocking the main loop.  On creation of new series it calls `tagMetric` every time (but then moves on and lets the http call complete asynchronously), on updates you can configure a an approximate interval (in number of updates) between calls to `tagMetric` (only really needed to make the tagDB self-healing and ensure consistency).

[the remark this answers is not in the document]
iksaif: can we add `state.database.tag()` which default implementation is `tagMetric()` ?
This way it's easy to override (for BigGraphite for example), but still does what you want for people who don't override it.

DanCech:
@iksaif I think all your concerns have been addressed at this point 😀 

@deniszh do you have any feedback?

lib/carbon/database.py
```diff
@@ -54,6 +59,26 @@ def validateArchiveList(self, archiveList):
     "Validate that the database can handle the given archiveList."
     pass
 
+  def tag(self, metric, httpRequest=None):
+    log.msg("Tagging %s" % metric)
```
iksaif: Add docstring here

lib/carbon/database.py
```diff
@@ -54,6 +59,26 @@ def validateArchiveList(self, archiveList):
...
+      return None
+
+    def successHandler(result, *args, **kw):
+      log.msg("Tagged %s: %s in %s" % (metric, result, time.time() - t))
+      return result
+
+    def errorHandler(err):
+      log.msg("Error tagging %s: %s" % (metric, err))
+
+    return httpRequest(
```
iksaif: Maybe don't return anything here ?

lib/carbon/util.py
```diff
@@ -298,3 +315,71 @@ def __init__(classObj, name, bases, members):
...
+  def format(tags):
+    return tags.get('name', '') + ''.join(sorted([
+      ';%s=%s' % (tag, value)
+      for tag, value in tags.items()
+      if tag != 'name'
+    ]))
+
+  @staticmethod
+  def encode(metric, sep='.'):
+    # if metric is tagged, encode it for storage in whisper etc
```
iksaif: Add an example of such encoding here ? (to make it clearer what the function does)

lib/carbon/util.py
```diff
@@ -298,3 +315,71 @@ def __init__(classObj, name, bases, members):
...
+
+class TaggedSeries(object):
+  @classmethod
+  def parse(cls, path):
+    tags = {}
+
+    # if path is in openmetrics format: metric{tag="value",...}
+    # https://github.com/RichiH/OpenMetrics
+    if path[-2:] == '"}' and '{' in path:
+      (metric, rawtags) = path[0:-1].split('{', 2)
```
iksaif: Same comment I made in graphite-web

lib/carbon/database.py
```diff
@@ -54,6 +59,26 @@ def validateArchiveList(self, archiveList):
     "Validate that the database can handle the given archiveList."
     pass
 
+  def tag(self, metric, httpRequest=None):
```
iksaif: I don't think this function needs to take httpRequest in argument (it only depends on reactor, which is global)

lib/carbon/writer.py
```diff
@@ -42,121 +54,174 @@
...
-    if not dataWritten:
-      time.sleep(1)
+    try:
+      t1 = time.time()
+      # If we have duplicated points, always pick the last. update_many()
+      # has no guaranted behavior for that, and in fact the current implementation
+      # will keep the first point in the list.
+      datapoints = dict(datapoints).items()
+      state.database.write(metric, datapoints)
+      if random.randint(1, settings.TAG_UPDATE_INTERVAL) == 1:  # nosec
```
iksaif: Isn't that a big usability issue ? It means that if you only send a few points to a metric it might not be created ? Or maybe I fail to understand why we need this here (add a comment in the code ?)

DanCech: There are 2 branches here, in the "new metric" branch `state.database.tag` is called unconditionally.  This branch is handling new data points for an existing series, it calls `state.database.tag` roughly every `settings.TAG_UPDATE_INTERVAL` points just to make sure that after upgrade, switching tag databases, or if there is an issue with the tagdb then the system will self-heal.

It would also be possible in the future for the tagdb to keep track of approximate first/last seen timestamps for each series so we can pass the time range through to the tagdb when searching via `seriesByTag`, but I have left that for later to keep things simple in this initial version.

lib/carbon/database.py
DanCech: Is there a reason not to return it? Right now writer.py doesn't do anything with it but in future maybe it would be useful

iksaif: Because it means that the API needs to return an HTTP request if you do that. Which clearly won't be the case in all implementations

lib/carbon/writer.py
iksaif: OK

DanCech:
Ok, @iksaif @deniszh @obfuscurity @cbowman0 I think everything brought up so far has been addressed.  Can I get a show of hands in favor of merging this beast?

Issue #676, opened 2017-07-28
rewrite-rule is being applied twice, resulting in double replacements


kalebwalton:
Example rewrite-rules.conf:

```
[post]
^(.*) = newprefix.\1
```

Results in the following metric: _newprefix.newprefix.mymetric_

I see that in service.py a pipeline is being created that applies rewrite:pre and rewrite:post rules, and it also includes an 'aggregate' pipe. The aggregate processor in lib/carbon/aggregator/processor.py also performs pre/post rewrites, so it's doubling them up. Submitting a PR to remove the rewrites from processor.py, although I am unsure if that's the most accurate solution.

deniszh:
My knowledge of rewriter is scarce, I'm not even sure does it a bug or not.

ganiserb:
This issue is also happening to me with the only rule:

```
[pre]
^(.+) = somePrefix.someOther.\1
```

The whisper files created result in something like this:
/var/lib/graphite/whisper/somePrefix/someOther/somePrefix/someOther/app1/...

PR #677, opened 2017-07-28
Remove pre and post rewrites from agg processor

kalebwalton:
Related to https://github.com/graphite-project/carbon/issues/676

### DIFF
--- a/lib/carbon/aggregator/processor.py
+++ b/lib/carbon/aggregator/processor.py
@@ -13,9 +13,6 @@ class AggregationProcessor(Processor):
   def process(self, metric, datapoint):
     increment('datapointsReceived')
 
-    for rule in RewriteRuleManager.rules(PRE):
-      metric = rule.apply(metric)
-
     aggregate_metrics = set()
 
     for rule in RuleManager.rules:
@@ -32,9 +29,6 @@ class AggregationProcessor(Processor):
         values_buffer.configure_aggregation(rule.frequency, rule.aggregation_func)
 
       values_buffer.input(datapoint)
-
-    for rule in RewriteRuleManager.rules(POST):
-      metric = rule.apply(metric)
 
     if settings.FORWARD_ALL and metric not in aggregate_metrics:
       if settings.LOG_AGGREGATOR_MISSES and len(aggregate_metrics) == 0:


deniszh:
Not sure if the fix is right, or even if a problem exists at all - but it breaks all aggregator's tests for sure.

kalebwalton:
You're right @deniszh - the fix is bad, however, there does appear to be a problem. The flow seems to go like this:

1) in service.py the createAggregatorService method creates a pipeline with 'rewrite:pre', 'aggregate', 'rewrite:post', 'relay' . The first 'rewrite' pipeline loads the rules into the RewriteRuleManager.
2) when a metric comes in it goes into the 'rewrite:pre' pipe, which is a RewriteProcessor first, which performs a rule execution. Then it goes into the 'aggregate' pipe, which is an AggregatorProcssor, which also performs a rule exection (duplicating it).

I don't know the right way to make it only execute once...

kalebwalton:
If I comment out the rule application lines from the RewriteProcessor then it doesn't duplicate rule executions.

```class RewriteProcessor(Processor):
  plugin_name = 'rewrite'

  def __init__(self, ruleset):
    self.ruleset = ruleset

  def process(self, metric, datapoint):
    #for rule in RewriteRuleManager.rules(self.ruleset):
    #  metric = rule.apply(metric)
    yield (metric, datapoint)```

**Comment by kalebwalton** [comment]
And just to point out the code that duplicates the rule execution (look at the RewriteRuleManager.rules(PRE) and (POST) areas):
```
class AggregationProcessor(Processor):
  plugin_name = 'aggregate'

  def process(self, metric, datapoint):
    increment('datapointsReceived')

    for rule in RewriteRuleManager.rules(PRE):
      metric = rule.apply(metric)

    aggregate_metrics = set()

    for rule in RuleManager.rules:
      aggregate_metric = rule.get_aggregate_metric(metric)

      if aggregate_metric is None:
        continue
      else:
        aggregate_metrics.add(aggregate_metric)

      values_buffer = BufferManager.get_buffer(aggregate_metric)

      if not values_buffer.configured:
        values_buffer.configure_aggregation(rule.frequency, rule.aggregation_func)

      values_buffer.input(datapoint)

    for rule in RewriteRuleManager.rules(POST):
      metric = rule.apply(metric)

    if settings.FORWARD_ALL and metric not in aggregate_metrics:
      if settings.LOG_AGGREGATOR_MISSES and len(aggregate_metrics) == 0:
        log.msg("Couldn't match metric %s with any aggregation rule. Passing on un-aggregated." % metric)
      yield (metric, datapoint)
```

PR #692, opened 2017-10-14

deniszh:
Rebase of #231, also see #580 and #437

Motivation:
1.) Avoid a large a number of small disk-writes.
2.) A more predictable write pattern across multiple carbon-cache instances i.e. every carbon-cache instance getting a share of disk I/O.
Modifications to meet the above points:
Modified writer.py to make carbon-cache do a whisper.update_many() only for those metrics which have more than MIN_DATAPOINTS_PER_UPDATE datapoints in cache. And this constraint is removed when the cache size grows above LOW_CACHE_SIZE, and comes back into effect when cache size shrinks below LOW_CACHE_SIZE.
LOW_CACHE_SIZE is set to 80% of MAX_CACHE_SIZE by default.


deniszh:
Please note that's backport of quite old PR, it's experimental and not tested - it's here for discussion mostly.

lib/carbon/cache.py
```diff
@@ -80,23 +80,38 @@ class MaxStrategy(DrainStrategy):
...
 
 class RandomStrategy(DrainStrategy):
   """Pop points randomly"""
-  def choose_item(self):
-    return choice(list(self.cache.keys()))  # nosec
-
+  def choose_item(self, mdpu=False):
+    metric_name = choice(list(self.cache.keys()))  # nosec
+    if mdpu:
+      count = int(0.1 * len(self.cache))
```
deniszh: TBH have no idea why taking 10% of whole length

lib/carbon/cache.py
```diff
@@ -71,7 +71,7 @@ def _generate_queue():
 
     self.queue = _generate_queue()
 
-  def choose_item(self):
+  def choose_item(self, mdpu=False):
```
deniszh: Now Naive strategy completely ignores MDPU. Not sure if it's fine...

lib/carbon/writer.py
```diff
@@ -57,12 +62,21 @@ def writeCachedDataPoints():
...
+      mdpu = False
+    else:
+      mdpu = True
+    (metric, datapoints) = cache.drain_metric(mdpu)
     if metric is None:
       # end the loop
       break
-
-    dbFileExists = state.database.exists(metric)
+    if state.cacheTooFull and MetricCache.size < settings.CACHE_SIZE_LOW_WATERMARK:
```
deniszh: Looks like it's deprecated now.  cache.drain_meric() will call cache.pop() which has this check already - https://github.com/graphite-project/carbon/blob/9fcb6f2c2acb6361a7461880f2373b085e5c42a2/lib/carbon/cache.py#L201

lib/carbon/writer.py
```diff
@@ -57,12 +62,21 @@ def writeCachedDataPoints():
...
+        events.cacheSpaceAvailable()
+    try:
```
deniszh: Also not sure why try..except needed here.

lib/carbon/cache.py
```diff
@@ -106,29 +121,40 @@ def _generate_queue():
...
+    self.mdpu_flag = False
 
     def _generate_queue():
       while True:
         t = time.time()
         metric_lw = sorted(self.cache.watermarks, key=lambda x: x[1], reverse=True)
         if settings.MIN_TIMESTAMP_LAG:
-          metric_lw = [x for x in metric_lw if t - x[1] > settings.MIN_TIMESTAMP_LAG]
+          metric_lw = [x for x in metric_lw if t - x[1] > settings.MIN_TIMESTAMP_LAG
+                        or (self.mdpu_flag and x[1] >= settings.MIN_DATAPOINTS_PER_UPDATE)]
```
deniszh: Ah, that's should be fixed too - with default MIN_DATAPOINTS_PER_UPDATE == 1 this effectively turns timesorted strategy to sorted.

deniszh:
I'm awaiting some feedback from @Exocomp and/or other interested parties, then probably make proper implementation for a couple of strategies.

Exocomp:
@deniszh 

> ...with default MIN_DATAPOINTS_PER_UPDATE == 1 this effectively turns timesorted strategy to sorted

Personally I'm interested in it being implemented for timesorted strategy, that would be more beneficial to me.

deniszh:
@Exocomp : OK, I removed all obscure changes - now it will work with timesorted strategy only, if you set MIN_DATAPOINTS_PER_UPDATE to some greater then 0 value.
Should work now, please try.

Exocomp:
@deniszh That's great, however I don't have a dev/test instance setup to easily test :(...

lib/carbon/cache.py
```diff
@@ -112,23 +112,25 @@ def _generate_queue():
...
+    self.mdpu_flag = False
 
     def _generate_queue():
       while True:
         t = time.time()
         metric_lw = sorted(self.cache.watermarks, key=lambda x: x[1], reverse=True)
         if settings.MIN_TIMESTAMP_LAG:
-          metric_lw = [x for x in metric_lw if t - x[1] > settings.MIN_TIMESTAMP_LAG]
+          metric_lw = [x for x in metric_lw if t - x[1] > settings.MIN_TIMESTAMP_LAG
+                        or (self.mdpu_flag and x[1] >= settings.MIN_DATAPOINTS_PER_UPDATE >0)]
```
DanCech: Can you explain what's going on here?

deniszh: Riiight, I probably screwed up things here. I should use cache.counts() instead of watermarks.  Back to drawing board!

lib/carbon/cache.py
```diff
@@ -112,23 +112,28 @@ def _generate_queue():
...
     def _generate_queue():
       while True:
         t = time.time()
         metric_lw = sorted(self.cache.watermarks, key=lambda x: x[1], reverse=True)
         if settings.MIN_TIMESTAMP_LAG:
           metric_lw = [x for x in metric_lw if t - x[1] > settings.MIN_TIMESTAMP_LAG]
+        if self.mdpu_flag and settings.MIN_DATAPOINTS_PER_UPDATE > 0:
+          metric_counts = [x for x in sorted(self.cache.counts, key=lambda x: x[1])
+                           if x[1] >= settings.MIN_DATAPOINTS_PER_UPDATE]
+          metric_lw.append(metric_counts)
```
DanCech: Here don't we want the intersection of the 2 lists? otherwise anything that has passed the min timestamp lag will be written regardless of the number of points.  For that to work I'd change both list comprehensions to return `x[0] for x in`, then you can use `metric_lw = list(set(metric_lw) & set(metric_counts))`, and remove the use of `itemgetter(0)` on line 144.

deniszh: That's a good quiestion. @Exocomp asked for "greedy" mechanics here - but now I'm wondering does it have any sense...

deniszh: Should work now, but now I'm afraid that's could be quite slow. Anyway, I don't think that we should merge this w/o thorough test under load, which is quite non-trivial.

DanCech: actually we could speed it up with:
```
metric_lw = [x[0] for x in sorted(self.cache.counts, key=lambda x: x[1])
                  if x[1] >= settings.MIN_DATAPOINTS_PER_UPDATE and x[0] in metric_lw]
```

deniszh: Yes, but that way breaks an order of metric_lw list. :(

DanCech: ah ok, so we want to keep the metrics_lw order, then we have to have 2 statements:
```
metric_count = [x[0] for x in self.cache.counts if x[1] >= settings.MIN_DATAPOINTS_PER_UPDATE]
metric_lw = [x for x in metric_lw if x in metric_count]
```
no need to sort the cache counts since we don't care about the order just whether they're over the threshold

deniszh: Ah, indeed - that looks much better! Thanks!

Issue #693, opened 2017-10-24
Extensive log message amount without disable switch


Kenterfie:
lib/carbon/database.py use log.msg in tag function what cause a huge log come up without option to disable.

Also the tagging function himself use many single http calls what also ends up in many refused connections. Would be great to modify these behavior.

DanCech:
Switching the success messages to debug isn't too hard, I'll pop a PR for that.  If you are seeing a lot of messages like `Error tagging X` then you need to properly configure your `GRAPHITE_URL='http://127.0.0.1:8000'` setting to point to your graphite-web install.

As far as the connections, it should be using a pool of persistent connections and you can adjust the frequency of the updates if they're overloading your graphite server using the `TAG_UPDATE_INTERVAL=100` config setting.  To halve the number of calls to graphite increase it to `200`.

DanCech:
The referenced PR also includes a setting `ENABLE_TAGS` which you can set to `False` if you aren't using tags at all and want to disable it altogether.

deniszh:
Merged to master, will be included in next release.

PR #694, opened 2017-10-24
Tagdb logging

DanCech:
This PR includes 3 commits that address #693 :

The first commit changes the tagdb logging to use debug for success messages and improves the log entries on error.

The second commit updates the http client to no longer produce log messages like:
```
24/10/2017 11:03:03 :: [console] Starting factory <twisted.web.client._HTTP11ClientFactory instance at 0x7fd0c235f758>
```
The 3rd commit adds a config setting `ENABLE_TAGS` which is `True` by default but can be set to `False` to disable tag functionality altogether if it isn't being used.


conf/carbon.conf.example
```diff
@@ -305,6 +305,9 @@ WHISPER_FALLOCATE_CREATE = True
 # URL of graphite-web instance, this is used to add incoming series to the tag database
 # GRAPHITE_URL = http://127.0.0.1:8000
 
+# Tag support, when enabled carbon will make HTTP calls to graphite-web to update the tag index
+# ENABLE_TAGS = True
```
deniszh: Very 👍

PR #696, opened 2017-11-08
Percentile aggregation methods

wridgers:
Percentile aggregation methods would be very useful.

### DIFF
--- a/lib/carbon/aggregator/rules.py
+++ b/lib/carbon/aggregator/rules.py
@@ -161,6 +161,7 @@ def count(values):
   if values:
     return len(values)
 
+
 def percentile(factor):
   def func(values):
     if values:
@@ -177,18 +178,18 @@ def percentile(factor):
   return func
 
 AGGREGATION_METHODS = {
-  'sum'   : sum,
-  'avg'   : avg,
-  'min'   : min,
-  'max'   : max,
-  'p50'   : percentile(0.50),
-  'p75'   : percentile(0.75),
-  'p80'   : percentile(0.80),
-  'p90'   : percentile(0.90),
-  'p95'   : percentile(0.95),
-  'p99'   : percentile(0.99),
-  'p999'  : percentile(0.999),
-  'count' : count
+  'sum': sum,
+  'avg': avg,
+  'min': min,
+  'max': max,
+  'p50': percentile(0.50),
+  'p75': percentile(0.75),
+  'p80': percentile(0.80),
+  'p90': percentile(0.90),
+  'p95': percentile(0.95),
+  'p99': percentile(0.99),
+  'p999': percentile(0.999),
+  'count': count,
 }
 
 # Importable singleton
--- a/lib/carbon/aggregator/rules.py
+++ b/lib/carbon/aggregator/rules.py
@@ -1,5 +1,7 @@
 import re
 
+from math import floor, ceil
+
 from os.path import exists, getmtime
 from twisted.internet.task import LoopingCall
 from cachetools import TTLCache, LRUCache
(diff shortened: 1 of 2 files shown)


deniszh:
IIRC it's only for aggregator.

deniszh:
>VALUES = [4, 8, 15, 16, 23, 42]
>         examples = [
             ('p999', 41.905, ),
             ('p99', 41.05, ),
             ('p95', 37.25, ),
             ('p90', 32.5, ),
             ('p80', 23, ),
             ('p75', 21.25, ),
             ('p50', 15.5, ),

Hmm...
I'm not sure that percentiles should work like this. 
Opinions? /cc @DanCech @cbowman0 @iksaif ?

iksaif:
aggregators are super fragile. I'm not super in favor of adding that kind of feature here.

DanCech:
This seems like it's setting people up for failure, since the first thing that's going to happen is that the data is going to get re-aggregated either on the next aggregation or at query time.

https://www.vividcortex.com/blog/why-percentiles-dont-work-the-way-you-think

deniszh:
>aggregators are super fragile. I'm not super in favor of adding that kind of feature here (but could be convinced overwise)

Aggregators are generally sucking, agreed. OTOH I do not think that this code will bring more instability there.

deniszh:
> This seems like it's setting people up for failure, since the first thing that's going to happen is that the data is going to get re-aggregated either on the next aggregation or at query time.

I personally still not sure that we accept that too, but I don't really like approximation code above.

But for aggregation of percentiles - yes, everybody knows that's not possible - but we already doing that! E.g. instrumentation libraries generating percentiles, for example http://metrics.dropwizard.io doing that. And these values will be also "re-aggregated either on the next aggregation or at query time",  that's known Graphite limitation - https://grafana.com/blog/2016/03/03/25-graphite-grafana-and-statsd-gotchas/#aggregating.percentiles
I mean, there's no difference what's source of this percentile is - aggregator or metric library, but people are still using metric libraries, why not using aggregator then?

iksaif:
I'd rather have buckets pushed into graphite with come kind of convention (using a format similar to prometheus). Plus a function to synthetize the buckets into percentiles. Some of that was done in https://github.com/graphite-project/graphite-web/pull/1571

wridgers:
Many people, include myself, are already doing percentile aggregation - but outside of graphite. Even though they are limited they are very useful. I don't think they should be excluded because someone could misunderstand or misuse them - that is true of many features.

wridgers:
> This seems like it's setting people up for failure, since the first thing that's going to happen is that the data is going to get re-aggregated either on the next aggregation or at query time.

This is already true of `avg`. If you send two samples in a minute that are aggregated by `avg` and twenty samples in the next minute, the re-aggregated metrics will be wrong.

For me, the link @deniszh posted sums up my feelings well. 

>  taking sensible shortcuts instead of obsessing over math is often the better way to accomplish your job of operating your infrastructure or software

I think percentiles are sufficiently useful that they should be available in the aggregators.

deniszh:
I agreed with @iksaif that proper percentile support should be done using buckets. But currently, we don't have it. But I also agreed with @wridgers that current status of percentiles support is useful despite all downsides.
Will it be enough to document new aggregations in documentation - https://graphite.readthedocs.io/en/latest/config-carbon.html#storage-aggregation-conf and also add some special warning about averaging percentiles (mentioning vividcortex' and/or Dieter's links)?
@DanCech @iksaif ?

iksaif:
sure

wridgers:
Great! :)

In that case, I'll put up a PR for the documentation this weekend.

deniszh:
Still have interest in that, @wridgers ?

wridgers:
> Still have interest in that, @wridgers ?

Ah yes. Christmas/life got in the way. It'd be great to get this merged. I'll look at this today!

wridgers:
I've rebased, fixed the linting errors, and put a documentation PR up here https://github.com/graphite-project/graphite-web/pull/2204.

PR #701, opened 2017-11-24
Added `--logger` and `--logfile` option.

zsuzhengdu:
Enable `--logger` and `--logfile` option offered by Twistd

```
  --logfile=LOGFILE     Log to a specified file, - for stdout
  --logger=LOGGER       A fully-qualified name to a log observer factory to
                        use for the initial log observer. Takes precedence
                        over --logfile and --syslog (when available).
```

### DIFF
--- a/lib/carbon/conf.py
+++ b/lib/carbon/conf.py
@@ -504,7 +504,15 @@ def get_default_parser(usage="%prog [options] <start|stop|status>"):
         "--instance",
         default='a',
         help="Manage a specific carbon instance")
-
+    parser.add_option(
+        "--logfile",
+        default='None',
+        help="Log to a specified file, - for stdout")
+    parser.add_option(
+        "--logger",
+        default='None',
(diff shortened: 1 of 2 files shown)


lib/carbon/conf.py
```diff
@@ -504,6 +504,10 @@ def get_default_parser(usage="%prog [options] <start|stop|status>"):
         "--instance",
         default='a',
         help="Manage a specific carbon instance")
+    parser.add_option(
+        "--logger",
+        default='None',
+        help="Use a log observer factory for the initial log observer")
```
DanCech: Can we use the same text as in the PR description here?  Also does it make sense to add support for `--syslog` and `--logfile` as well for completeness?

zsuzhengdu: `--syslog` is already supported https://github.com/graphite-project/carbon/blob/master/lib/carbon/conf.py#L470

Added support for `--logfile` as well

deniszh:
Looks good, @DanCech ?

PR #708, opened 2017-12-04
add codecov badge

DanCech:


deniszh:
ah, `https://github.com/graphite-project/carbon/blob/master/bin/carbon-aggregator-cache.py#L33:8` should be fixed 🤕

DanCech:
@deniszh yup, sorted out the lint issues in bin also 😉

PR #710, opened 2017-12-12
restore droppedCreates metric

fessyfoo:
the refactor in 88a6464 accidentally stopped incremeting dropedCreates

note i didn't test this yet, if you like consider it a bug report instead of a PR. :)


deniszh:
I'm not sure that 100% matching old metric. But OTOH - it's better than nothing. LGTM.

PR #713, opened 2017-12-15
Py3 queue

piotr1212:
impor queue from six.moves for py2/3 compat


DanCech:
Looks good @piotr1212 what's the reason for the benchmark_cache change?

piotr1212:
Takes more than 120 seconds on a latest generation i7, seems a bit long to me. Can revert if you prefer.

DanCech:
Ah, ok

Issue #714, opened 2017-12-21
Carbon fnv1a_ch not compatible with carbon-c-relay


iain-buclaw-sociomantic:
Given the following carbon config:
```
DESTINATIONS = 192.168.240.186:2003:a, 192.168.240.184:2003:b
HASHING_TYPE = fnv1a_ch
```
And the equivalent carbon-c-relay config:
```
cluster graphite
    fnv1a_ch
        192.168.240.186:2003=a proto tcp
        192.168.240.184:2003=b proto tcp
    ;
```

They both produce different hash rings.

Carbon:
```
  218@192.168.240.184:2003   536@192.168.240.184:2003   611@192.168.240.186:2003
  991@192.168.240.184:2003  1625@192.168.240.186:2003  2470@192.168.240.186:2003
 2505@192.168.240.184:2003  2801@192.168.240.184:2003  3088@192.168.240.184:2003
 3112@192.168.240.186:2003  3378@192.168.240.186:2003  3423@192.168.240.186:2003
 3696@192.168.240.186:2003  3859@192.168.240.184:2003  4185@192.168.240.184:2003
 4496@192.168.240.186:2003  5319@192.168.240.184:2003  6103@192.168.240.184:2003
 6265@192.168.240.184:2003  6314@192.168.240.184:2003  6566@192.168.240.184:2003
 6755@192.168.240.186:2003  7335@192.168.240.186:2003  7474@192.168.240.184:2003
 7671@192.168.240.184:2003  7877@192.168.240.186:2003  7948@192.168.240.186:2003
 8078@192.168.240.186:2003  9040@192.168.240.184:2003  9314@192.168.240.186:2003
 9796@192.168.240.184:2003 10098@192.168.240.184:2003 10427@192.168.240.186:2003
11346@192.168.240.186:2003 11432@192.168.240.186:2003 12229@192.168.240.184:2003
12313@192.168.240.184:2003 12612@192.168.240.184:2003 12684@192.168.240.186:2003
12862@192.168.240.186:2003 12941@192.168.240.186:2003 12973@192.168.240.186:2003
13406@192.168.240.184:2003 14002@192.168.240.186:2003 14373@192.168.240.184:2003
14550@192.168.240.186:2003 14711@192.168.240.184:2003 14715@192.168.240.184:2003
15042@192.168.240.186:2003 15100@192.168.240.186:2003 15303@192.168.240.184:2003
[...]

PR #717, opened 2017-12-23
Fix hash position computation of fnv1a_ch

DanCech:
This PR is an attempt to future-proof the fnv1a_ch hash computation by properly handling multi-byte characters in addition to applying the fix described in #714 and graphite-project/graphite-web#2165 & graphite-project/graphite-web#2168.

The way we use the fnv32a hash function is now consistent with other hash functions where any string is encoded into utf-8 before being hashed. This produces consistent results across python 2.x and 3.x with or without the pyhash library, and should be consistent with any third-party implementations.

This PR also syncs hashing.py between carbon and graphite-web to ensure they produce consistent results.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,18 @@ matrix:
         - TOXENV=py36
     - python: 3.6
       env:
+        - TOXENV=py36-pyhash
+    - python: 3.6
+      env:
         - TOXENV=lint
 
 env:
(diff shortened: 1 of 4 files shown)


deniszh:
Will backport this to 1.1.x, and release 1.1.2, but I think in 2018 - let's wait for some other fixes.

PR #722, opened 2018-01-08
Add TCP_KEEPALIVE

iksaif:
The RATIO_RESET settings are not always robust enough to detect
all TCP issues. Also add the possibility to enable TCP KeepAlive
which is design to detect broken connections.


lib/carbon/util.py
```diff
@@ -33,6 +33,11 @@
   import pickle
   USING_CPICKLE = False
 
+import socket
+from time import sleep, time
```
deniszh: time already imported

lib/carbon/util.py
```diff
@@ -33,6 +33,11 @@
...
+from twisted.python.util import initgroups
```
deniszh: already imported in line 15

lib/carbon/util.py
```diff
@@ -33,6 +33,11 @@
...
+from twisted.scripts.twistd import runApp
```
deniszh: same here

deniszh:
Lint is also very angry

lib/carbon/util.py
```diff
@@ -42,6 +43,25 @@ def dropprivs(user):
   return (uid, gid)
 
 
+def enableTcpKeepAlive(transport):
+  from carbon.conf import settings
+
+  if not settings.TCP_KEEPALIVE or not hasattr(transport, 'getHandle'):
```
DanCech: I'd prefer to see the setting passed in when the function is called, that way it's clearer what's going on.

iksaif: or can I name it `maybeEnableTcpKeepAlive()` and pass it `settings` ? Less code duplication

DanCech: I would prefer to just pass the setting itself, not a big fan of the `maybe` pattern

iksaif: done

DanCech:
Apologies, I didn't see that settings is also used for KEEPINT etc, in that case it'll be cleaner to pass in settings directly I guess.

iksaif:
The new version takes 'enabled' + additional settings, so it's slightly better than better.

PR #723, opened 2018-01-08
client: make sure the line client doesn't loose precision

iksaif:
For example '1.498566361088E12' would be printed as '1.49856636109e+12'
with the old version.

### DIFF
--- a/lib/carbon/client.py
+++ b/lib/carbon/client.py
@@ -413,7 +413,7 @@ class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
 
   def _sendDatapointsNow(self, datapoints):
     for metric, datapoint in datapoints:
-        self.sendLine("%s %s %d" % (metric, datapoint[1], datapoint[0]))
+        self.sendLine("%s %f %d" % (metric, datapoint[1], datapoint[0]))
 
 
 class CarbonLineClientFactory(CarbonClientFactory):
--- a/lib/carbon/tests/test_client.py
+++ b/lib/carbon/tests/test_client.py
@@ -72,7 +72,7 @@ class CarbonLineClientProtocolTest(TestCase):
 
   def test_send_datapoints_now(self):
     datapoint = ('foo.bar', (1000000000, 1.0))
-    expected_line_to_send = "foo.bar 1.0 1000000000"
+    expected_line_to_send = "foo.bar 1.000000 1000000000"
 
     self.protocol._sendDatapointsNow([datapoint])
     self.protocol.sendLine.assert_called_once_with(expected_line_to_send)


lib/carbon/client.py
```diff
@@ -413,7 +413,7 @@ class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
 
   def _sendDatapointsNow(self, datapoints):
     for metric, datapoint in datapoints:
-        self.sendLine("%s %s %d" % (metric, datapoint[1], datapoint[0]))
+        self.sendLine("%s %f %d" % (metric, datapoint[1], datapoint[0]))
```
deniszh: Hmmm.... Would it convert all values to float explicitly?

iksaif: I think all values are casted to float anyway in the code

lib/carbon/tests/test_client.py
```diff
@@ -72,7 +72,7 @@ def setUp(self):
 
   def test_send_datapoints_now(self):
     datapoint = ('foo.bar', (1000000000, 1.0))
-    expected_line_to_send = "foo.bar 1.0 1000000000"
+    expected_line_to_send = "foo.bar 1.000000 1000000000"
```
DanCech: This is going to have a negative impact on bandwidth usage, we should be sending the smallest possible representation.

iksaif: I can't find anything smarter than `('%f' % x).rstrip('0').rstrip('.')`, any idea ?

iain-buclaw-sociomantic:
> For example '1.498566361088E12' would be printed as '1.49856636109e+12' with the old version.

Shouldn't there be a test for this?

DanCech:
Yes, there should.  From what I can make out the use of `%f` has a number of drawbacks vs `%s`, it forces all numeric values to be represented with 6 decimal places of precision.  In any case the difference between 1.498566361088E12 and 1.49856636109e+12 is 0.0000000001% so if that's really a concern send the value as 1498566361088 which will be represented exactly.

IMHO this whole patch should simply be reverted.

@iksaif @deniszh thoughts?

deniszh:
Yes, I think we should revert it, or replace with something else. Not sure how good `('%f' % x).rstrip('0').rstrip('.')` will work.

DanCech:
There's no need to change the behavior from before this patch that I can see.  `('%f' % x).rstrip('0').rstrip('.')` seems like it will cause more problems than it solves.

iksaif:
This can't really be reverted as numbers are messed up with %s. For example if you try to send a timestamp in millisecond (which is pretty common) you'll simply get the wrong value. So yes, it's 0.0000000001%, has real world effects.

deniszh:
>For example if you try to send a timestamp in millisecond (which is pretty common, we found almost a thousand places doing that) you'll simply get the wrong value.

Hmmmm.... I do not see any problems with ms timestamps.
How was it working before then? I'm puzzled :|

python 2.7
```
>>> import time
>>> now_ms = time.time() * 1000
>>> expected_line_to_send = "foo.bar %f 1000000000" % now_ms
>>> old_line_to_send = "foo.bar %d 1000000000" % now_ms
>>> print old_line_to_send
foo.bar 1516609859926 1000000000
>>> print expected_line_to_send
foo.bar 1516609859926.515137 1000000000
>>> quit
Use quit() or Ctrl-D (i.e. EOF) to exit
>>> quit
```
python 3.6
>>> import time
>>> now_ms = time.time() * 1000
>>> expected_line_to_send = "foo.bar %f 1000000000" % now_ms
>>> old_line_to_send = "foo.bar %d 1000000000" % now_ms
>>> old_line_to_send = "foo.bar %d 1000000000" % now_ms
foo.bar 1516610031086 1000000000
>>> print(expected_line_to_send)
foo.bar 1516610031086.135986 1000000000
>>> quit
Use quit() or Ctrl-D (i.e. EOF) to exit
```


>Why do you think the additional '0' are really an issue ? (We've found out that the bottleneck is more often CPU than bandwidth).

It's not a single '0', it's '.000000' - 7 bytes more, for every datapoint.
People are complaining about a bandwith too, and CPU is much easier to scale than a network.

**Comment by iksaif** [comment]
It was never working with the line client before. Between carbon instances, pickle is still the default, you can use the line protocol with `DESTINATION_PROTOCOL = line`.

**Comment by deniszh** [comment]
Ah, got it. Then it's mostly safe to enable it, right, @DanCech ?
But I can't reproduce a problem btw, could you please check what I'm doing wrong above, @iksaif ?

**Comment by iksaif** [comment]
The is is with `%s` not `%d` (and you can't use `%d` either because the value is supposed to be a float)

```
>>> a = 0.5
>>> "%d %s %f" % (a, a, a)
'0 0.5 0.500000'
>>> now_ms = time.time() * 1000
>>> "%d %s %f" % (now_ms, now_ms, now_ms)
'1516618449319 1.51661844932e+12 1516618449319.762939'
```

deniszh:
Right. Well, that's for optional proto and having `%s` is definitely wrong there. I'm not a big fan of `%f` there, but it can stay for now as we have no other ideas yet.

Issue #730, opened 2018-01-10
Memory leak since upgrading to carbon 1.1.1


wridgers:
Looks like a memory leak to me. Cache size hasn't changed, but the memory is growing.

![2018-01-10-131405_1261x270_scrot](https://user-images.githubusercontent.com/235730/34774778-9fa65bda-f608-11e7-8cb3-1da1004e7cba.png)
![2018-01-10-131400_838x267_scrot](https://user-images.githubusercontent.com/235730/34774785-a293f1b8-f608-11e7-8914-0f7147088064.png)

DanCech:
Did you set GRAPHITE_URL in your carbon config?  If not it may be due to it being unable to update the tag db

wridgers:
`GRAPHITE_URL` is set in `carbon.conf` to `http://127.0.0.1:80/`.

DanCech:
Is that a valid url for your graphite-web install?

wridgers:
> Is that a valid url for your graphite-web install?

`curl -s http://127.0.0.1:80` returns a PNG, so I'm pretty sure it's correct.

wridgers:
I'm not seeing any errors in the logs about an invalid URL and the `tags` tables in the database (postgres) are being populated.

DanCech:
Folding this into #729

deniszh:
Looks like different issue, reopening.

deniszh:
Thanks, @wridgers !
Fix' merged, will be released soon.

Issue #729, opened 2018-01-10
Random gap in recent data since upgrading to 1.1.1


wridgers:
Since upgrading to carbon 1.1.1 I'm seeing a gap in recent metrics. The data doesn't seem to be "missing" - it does eventually get written to disk. There seems to be a 10m gap in the last 20m at all times.

![2018-01-10-110846_845x294_scrot](https://user-images.githubusercontent.com/235730/34769840-ed40ba14-f5f6-11e7-9837-66a024556939.png)
![2018-01-10-110906_1261x379_scrot](https://user-images.githubusercontent.com/235730/34769843-eeb2fd6c-f5f6-11e7-8d45-00be952982d2.png)

Sometimes the last 10m of data is missing, and then all my grafana alerts will fire and complain about no data.

deniszh:
Is it only for carbon (internal) metrics or all metrics?

wridgers:
All metrics. :(

DanCech:
Sounds like maybe whisper is lagging 20m behind but the most-recent data is getting served by carbonlink.  Do you have your graphite url set in the carbon config?

deniszh:
>Sounds like maybe whisper is lagging 20m behind but the most-recent data is getting served by carbonlink.

In this case should be no gaps, if carbonlink merging works properly ofc...

wridgers:
We've set `ENABLE_TAGS=False` and restarted the caches. It looks like the problem is solved, but it'll take a while to confirm.

https://github.com/graphite-project/carbon/issues/730 Seems like might be fixed too.

Waiting on more data to come in.

deniszh:
#730 is clear, but non-configure Tags causing gaps... That's a bit worrying.

wridgers:
I diffed the config that I commited into version control, and there nothing but new lines with comments. :/

wridgers:
> We've set ENABLE_TAGS=False and restarted the caches. It looks like the problem is solved, but it'll take a while to confirm.
> Seems like #730 might be fixed too.
> Waiting on more data to come in.

It's not fixed - the gaps are back. Perhaps the restart just made the problem go away for a bit.

DanCech:
Ok, so if you have `ENABLE_TAGS=False` then it has to be something else.  Are you seeing memory usage increasing also?

wridgers:
> Are you seeing memory usage increasing also?

Yeah, memory usage of the caches is still growing while the number of metrics in the cache stays constant.

[...]

PR #733, opened 2018-01-12
Fix leaks in writeForever and writeTagsForever

cevans-uk:
`writeCachedDataPoints` will return a `Deferred` before it's finished executing (most noticeable if it needs to wait for a token from `CREATE_BUCKET` or `UPDATE_BUCKET`), so ignoring the return value before calling it again can result in lots of overlapping calls.

I have the `writeForever` patch running on a live system for about an hour, and it has fixed both #729 and #730 (which were both caused by lots of calls taking data from the cache but having to wait to write it to disk because of `MAX_UPDATES_PER_SECOND`).

We have tags disabled, so I'm not certain that `writeTags` has the same leak: I think it'd be fine if it never takes more than 0.2s (the sleep in `writeTagsForever`). I don't know if this is the case or not.

### DIFF
--- a/lib/carbon/writer.py
+++ b/lib/carbon/writer.py
@@ -183,10 +183,11 @@ def writeCachedDataPoints():
             pointCount, metric, updateTime))
 
 
+@inlineCallbacks
 def writeForever():
   while reactor.running:
     try:
-      writeCachedDataPoints()
+      yield writeCachedDataPoints()
     except Exception:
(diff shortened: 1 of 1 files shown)


DanCech:
Good stuff, thanks!

deniszh:
Yes, indeed! Thanks, @cevans-uk !
Merging and will backport to 1.1.x

PR #738, opened 2018-01-17
writers run synchronously in separate threads

DanCech:
This PR removes the async code from the writers, which now run synchronously in their own threads.  This should alleviate issues caused by mixing code that runs in the main thread with workers running in separate threads.

It does add a dependency on `urllib3` since we're no longer using the twisted http library.

### DIFF
--- a/lib/carbon/http.py
+++ b/lib/carbon/http.py
@@ -1,74 +1,21 @@
-from twisted.internet import defer, protocol, reactor
-from twisted.web import client
-from twisted.web.http_headers import Headers
-from twisted.web.iweb import IBodyProducer
-from zope.interface import implements
-
-try:
-  from urllib import urlencode
-except ImportError:
-  from urllib.parse import urlencode
-
-
-class QuietHTTP11ClientFactory(client._HTTP11ClientFactory):
-    noisy = False
-
+import urllib3
 
 # async http client connection pool
-pool = client.HTTPConnectionPool(reactor)
-pool._factory = QuietHTTP11ClientFactory
+http = urllib3.PoolManager()
 
 
-class StringProducer(object):
-  implements(IBodyProducer)
+def httpRequest(url, values=None, headers=None, method='POST', timeout=5):
+  try:
+    result = http.request(
+      method,
+      url,
+      fields=values,
+      headers=headers,
+      timeout=timeout)
+  except BaseException as err:
+    raise Exception("Error requesting %s: %s" % (url, err))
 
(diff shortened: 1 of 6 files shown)


lib/carbon/http.py
```diff
@@ -1,74 +1,21 @@
...
+http = urllib3.PoolManager()
 
 
-def httpRequest(url, values=None, headers=None, method='POST'):
-  fullHeaders = {
-    'Content-Type': ['application/x-www-form-urlencoded']
-  }
-  if headers:
-    fullHeaders.update(headers)
+def httpRequest(url, values=None, headers=None, method='POST', timeout=5):
```
deniszh: i'm wondering how it will affect performance, especially when GRAPHITE_URL is not configured or configured wrong. Will it block whole processing for 5 seconds?

DanCech: It should just block the tag writer thread, I'll verify that.

lib/carbon/http.py
```diff
@@ -1,74 +1,21 @@
...
+      headers=headers,
+      timeout=timeout)
+  except BaseException as err:
+    raise Exception("Error requesting %s: %s" % (url, err))
 
-  def handle_response(response):
-    d = defer.Deferred()
-    response.deliverBody(SimpleReceiver(response, d))
-    return d
+  if result.status != 200:
```
iksaif: maybe >= 300 ?

requirements.txt
```diff
@@ -2,3 +2,4 @@ Twisted>=13.2.0
 git+git://github.com/graphite-project/whisper.git#egg=whisper
 txAMQP
 cachetools
+urllib3
```
iksaif: is there a rationale for urllib3 vs. request ? (curious)

lib/carbon/http.py
DanCech: Graphite will return 200, why do you think it should be more permissive?

requirements.txt
DanCech: Because we already use it in graphite-web

DanCech: fwiw, requests is a wrapper around urllib3

lib/carbon/http.py
DanCech: Verified that it just blocks the tag writer thread

deniszh: 👍

requirements.txt
deniszh: agreed, `requests` is just extra dependency here.

PR #743, opened 2018-02-01
Support for storing tagged series in hashed filenames

DanCech:
This PR adds support for naming tagged metric filenames using only the hash of the metric name, which avoids issues with filename length limits when multiple tags are used.

The hashed naming scheme is now the default, to return to the old scheme the user can add `TAG_HASH_FILENAMES = False` to their config.

It would theoretically be possible for the system to check if a file using the old name exists before creating a new file, that's something I'll look into.  I also want to do more testing.


deniszh:
> It would theoretically be possible for the system to check if a file using the old name exists before creating a new file, that's something I'll look into. I also want to do more testing.

Yep, that could be nice, indeed - otherwise migration to new scheme is not really possible without loosing tagged metrics, right?

DanCech:
That's one of the things I need to test, in theory the old and new files would both be read and the data merged.

DanCech:
My last commit handles switching from non-hashed to hashed filenames more gracefully.  The `exists` function in the databases will first check for the expected file, if it's not found and hashed filenames are enabled then it will check for a non-hashed file and if found rename it to the hashed name.

This will resolve the issue of doubling disk usage when enabling hashed filenames and avoid the overhead of graphite-web having to merge the files together at run-time.

deniszh:
👍

lib/carbon/database.py
```diff
@@ -123,7 +124,13 @@ def write(self, metric, datapoints):
       whisper.update_many(path, datapoints)
 
     def exists(self, metric):
-      return exists(self.getFilesystemPath(metric))
+      if exists(self.getFilesystemPath(metric)):
+        return True
+      # if we're using hashed filenames and a non-hashed file exists then move it to the new name
+      if self.tag_hash_filenames and exists(self.getFilesystemPath(metric, False)):
+        os.rename(self.getFilesystemPath(metric, False), self.getFilesystemPath(metric))
```
deniszh: it shouldn't rename same file in different threads, right? no race condition here?

DanCech: The write process runs in a single dedicated thread, the whole procedure where the writer first checks if the file exists and creates it if needed isn't thread safe today.

lib/carbon/database.py
```diff
@@ -51,7 +51,7 @@ def setMetadata(self, metric, key, value):
     "Modify metric metadata."
     raise NotImplementedError()
 
-  def getFilesystemPath(self, metric):
+  def getFilesystemPath(self, metric, tag_hash_filenames=None):
```
iksaif: Does it need to be exposed to the Database API ? Looks like a whisper specific thing to me.

DanCech: It's used by Ceres also, if it's a concern I could switch both Whisper and Ceres to use an internal _ version that takes the additional parameter

DanCech: I'll do that, it'll be cleaner that way

DanCech:
@iksaif that should solve the problem

PR #747, opened 2018-02-11
Changing default Graphite URL to http://127.0.0.1:80

deniszh:


lib/carbon/conf.py
```diff
@@ -76,7 +76,7 @@
   AMQP_VERBOSE=False,
   AMQP_SPEC=None,
   BIND_PATTERNS=['#'],
-  GRAPHITE_URL='http://127.0.0.1:8000',
+  GRAPHITE_URL='http://127.0.0.1:80',
```
iksaif: The default command line to start a dev server binds on 8080, wouldn't that make more sense ?

deniszh: Good question. 8080 is also good. @DanCech ?

DanCech: Not sure what the best solution is, since in most production deployments it's going to be at 80

deniszh: I would say let's aim for prod deployments here...

iksaif:
:+1:

Issue #766, opened 2018-04-16
[console] Unhandled Error


flowppp:
Hi All,

I have this error message when my carbon-relay is working : 

16/04/2018 09:34:28 :: [console] Unhandled Error
Traceback (most recent call last):
  File "/opt/programs/graphite/lib64/python3.4/site-packages/twisted/protocols/basic.py", line 454, in dataReceived
    self.lineReceived(line)
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/protocols.py", line 179, in lineReceived
    self.metricReceived(metric, datapoint)
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/protocols.py", line 157, in metricReceived
    events.metricReceived(metric, datapoint)
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/events.py", line 17, in __call__
    handler(*args, **kwargs)
--- <exception caught here> ---
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/pipeline.py", line 31, in run_pipeline
    for out_metric, out_datapoint in processor.process(metric, datapoint):
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/client.py", line 569, in process
    state.client_manager.sendDatapoint(metric, datapoint)
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/client.py", line 554, in sendDatapoint
    for destination in self.getDestinations(metric):
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/client.py", line 546, in getDestinations
    destinations = list(self.router.getDestinations(metric))
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/routers.py", line 108, in getDestinations
    for (server, instance) in self.ring.get_nodes(key):
  File "/opt/programs/graphite/lib64/python3.4/site-packages/carbon/hashing.py", line 108, in get_nodes
    index = bisect.bisect_left(self.ring, search_entry) % self.ring_len
builtins.TypeError: unorderable types: tuple() < NoneType()

I have absolutly no idea what's can be.

Thanks for your help
Flo

piotr1212:
Could you please post your carbon.conf?

flowppp:
i'am using python 3.4

piotr1212:
When does this happen? Is there some specific data (in the metricname) that triggers this?

flowppp:
[...]

PR #778, opened 2018-05-20
hashing, fix bisect on py3

piotr1212:
fixes: #766

### DIFF
--- a/lib/carbon/hashing.py
+++ b/lib/carbon/hashing.py
@@ -91,7 +91,7 @@ class ConsistentHashRing:
   def get_node(self, key):
     assert self.ring
     position = self.compute_ring_position(key)
-    search_entry = (position, None)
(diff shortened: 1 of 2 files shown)


DanCech:
Looks good at first glance, this code is duplicated in graphite-web so you'll want to open a PR there also.

piotr1212:
@DanCech What about removing the code from graphite-web and importing from carbon? is there some reason why that is not wanted except for web depending on carbon?

deniszh:
That's possible, but please note that carbon requires e.g. Twisted, so, it will bring more dependencies to graphite-web. Not the end of the world, though.

piotr1212:
Hmm, will leave it separate for now.

ccope:
There's a stray swp file in the commit

piotr1212:
removed the swp file

PR #790, opened 2018-07-13
rewrite is handled in pipeline now

### LINKED ISSUE #676: rewrite-rule is being applied twice, resulting in double replacements

DanCech:
This expands #677 by also removing the outdated tests and no-longer-needed imports.

Resolves #676

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # http://travis-ci.org/#!/graphite-project/carbon
 language: python
 python: 2.7
+# sudo is needed to install libboost-python-dev, which is required by pyhash
 sudo: required
 
 matrix:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 # http://travis-ci.org/#!/graphite-project/carbon
 language: python
 python: 2.7
-sudo: false
+sudo: required
```
iksaif: Please add a code comment explaining why

PR #793, opened 2018-08-12
Support two new config options to enable SSL transport.

postwait:
* DESTINATION_TRANSPORT = ssl (default "none")
 * DESTINATION_SSL_CA = /path/to/bundle.crt (default system budle)


postwait:
This acts the same way carbon-c-relay does... they can both talk to SSL/TLS
capable endpoints.

lib/carbon/client.py
```diff
@@ -2,8 +2,9 @@
 from time import time
 from six import with_metaclass
 
+from OpenSSL import SSL
```
deniszh: Will it make pyOpenSSL required dependency then? Could it be optional then?

lib/carbon/client.py
```diff
@@ -2,8 +2,9 @@
...
 from twisted.application.service import Service
-from twisted.internet import reactor
+from twisted.internet import ssl, reactor
```
deniszh: Will it work on Twisted version < 14 ? 
If not - then we should raise min Twisted version in requirements.txt

lib/carbon/client.py
postwait: I've updated to make it optional.

lib/carbon/client.py
postwait: It does, however it does not include the tlsext hostname prior to 14. So, you still get a secure connection, but that TLS extension is not used..

postwait:
What is the process for getting this integrated into master and then subsequent releases?

deniszh:
I can merge it already, but usually it's nice to have more than one pair of eyes to review. Looks like @DanCech  and @iksaif are still busy. Maybe @piotr1212  could review it?

postwait:
I am very supportive of more eyeballs. Python, nor twisted, is my usual programming environment.

deniszh:
Cool, thanks, @cbowman0 !
Merging.

PS: I'm back from vacation and planning to prepare next release soon.

piotr1212:
Already merged but my comments:

when opening a file in Python it is common to use a `context manager` (the `with` statement):
```
with open('path') as f:
    do_stuff(f.read())
```
File will be closed when you are done.

Why is the setting of tcp_keepalive disabled for SSL connections, is this not supported?

piotr1212:
```client = ssl.optionsForClientTLS(self.host.decode('utf-8'), authority)```
Is not Python 3 compatible. strings are a real headache to make compatible between versions.

In Py2 there are two types which represent strings, `str` which is a byte array and `unicode` which represents unicode strings. In Py3 they switched to unicode strings by default, making `str` a unicode string, `unicode` an alias to `str`. `bytes` is a byte array (py2 style `str`).

I would suggest using `six.u()` for this case:
```
from six import u
client = ssl.optionsForClientTLS(u(self.host), authority)`
```

deniszh:
Wow, thanks, @piotr1212 !
Sorry for not waiting your review, thought you was busy...

@postwait , could you please check issues above? Thanks!

PR #797, opened 2018-08-23
Add tests for ms timestamps

iksaif:


piotr1212:
I didn't know this is supported

DanCech:
I'm also not aware of any code in carbon itself that would support this, can you give us any more info @iksaif ?

iksaif:
Sorry for the lack of context, I've sent all the pending patches that I had and for some of them I also forgot why. I'll drop this one until I can find a good reason to send it back. I think it was to make sure the timestamps would not be changed during processing (whatever value they are), but not sure it really makes sense.

PR #800, opened 2018-08-23
Stop accept() when we reach MAX_RECEIVER_CONNECTIONS

iksaif:
Together with SO_REUSPORT this allow to load balance between
instances on the same machine.


iksaif:
Imagine that you have 10 carbon processes on the same machine listening on the same TCP port with SO_REUSEPORT. If you limit the number of accepted connections per client you'll enforce a fair distributions of clients per process. Additionally, when the machine is full, new connections will automatically be denied and go to machines with spare capacity.

PR #801, opened 2018-08-23
Add DESTINATIONS_POOL_REPLICAS

iksaif:
This allows to have multiple connections per destinations, this will
pool all the replicas of a single host in the same queue and distribute
points accross these replicas instead of replicating them.
The following example will balance the load between :0 and :1.
    
```
       DESTINATIONS = foo:2001:0, foo:2001:1
       RELAY_METHOD = rules
```


conf/carbon.conf.example
```diff
@@ -409,6 +409,15 @@ DESTINATIONS = 127.0.0.1:2004
 # DESTINATION_TRANSPORT = none
 # DESTINATION_SSL_CA=/path/to/private-ca.crt
 
+# This allows to have multiple connections per destinations, this will
+# pool all the replicas of a single host in the same queue and distribute
+# points accross these replicas instead of replicating them.
+# The following example will balance the load between :0 and :1.
+## DESTINATIONS = foo:2001:0, foo:2001:1
+## RELAY_METHOD = rules
+# Noet: this is currently incompatible with USE_RATIO_RESET which gets
```
deniszh: s/Noet/Note/

iksaif: fixed

PR #802, opened 2018-08-26
Fix compatibility issues 

deniszh:
Fixing issues mentioned by @piotr1212 in review for #793

### DIFF
--- a/lib/carbon/client.py
+++ b/lib/carbon/client.py
@@ -272,14 +272,15 @@ class CarbonClientFactory(with_metaclass(PluginRegistrar, ReconnectingClientFact
        authority = None
        if settings['DESTINATION_SSL_CA']:
            try:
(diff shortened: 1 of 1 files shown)


deniszh:
/cc @piotr1212 @postwait 
Piotr, you also mentioned that TCP Keepalive disabled for SSL connections, but I didn't find that in the code. Could you please elaborate?

piotr1212:
https://github.com/graphite-project/carbon/blob/65b102ce761fa629b5769b3a06857e55bb93b265/lib/carbon/util.py#L50-L67
So if it is an SSL connection, the keepalive flags are not set. I don't see a particular reason why not to enable keepalive with SSL, maybe it is is implicitly set somewhere in twisted or I'm just missing some reason.

deniszh:
Ah, OK. But it's disabled explicitly, so, OK for now, I think

Issue #805, opened 2018-08-31
Carbon doesn't start with _resolver issues


ybizeul:
I think 6990eaa broke some stuff but I can't figure out what's wrong yet.

```
Aug 31 16:05:46 nabox carbon-relay.py[13506]: File "/opt/graphite/lib/carbon/service.py", line 183, in setupRelayProcessor
Aug 31 16:05:46 nabox carbon-relay.py[13506]: from carbon.client import CarbonClientManager
Aug 31 16:05:46 nabox carbon-relay.py[13506]: File "/opt/graphite/lib/carbon/client.py", line 15, in <module>
Aug 31 16:05:46 nabox carbon-relay.py[13506]: from carbon.resolver import setUpRandomResolver
Aug 31 16:05:46 nabox carbon-relay.py[13506]: File "/opt/graphite/lib/carbon/resolver.py", line 5, in <module>
Aug 31 16:05:46 nabox carbon-relay.py[13506]: from twisted.internet._resolver import GAIResolver
Aug 31 16:05:46 nabox carbon-relay.py[13506]: ImportError: No module named _resolver
Aug 31 16:05:46 nabox systemd[1]: carbon-relay.service: control process exited, code=exited status=1
Aug 31 16:05:46 nabox systemd[1]: Failed to start graphite carbon relay.
```

```
# pip freeze
/usr/local/lib/python2.7/dist-packages/cffi/model.py:531: UserWarning: 'point_conversion_form_t' has no values explicitly defined; guessing that it is equivalent to 'unsigned int'
  % self._get_c_name())
attrs==18.1.0
Automat==0.7.0
boto3==1.8.5
botocore==1.11.5
cachetools==2.1.0
certifi==2018.8.24
cffi==1.11.5
characteristic==0.1.0
chardet==3.0.4
click==6.7
constantly==15.1.0
cryptography==0.6.1
Django==1.11.15
django-tagging==0.4.6
docutils==0.14
elasticsearch==6.3.1
elasticsearch-curator==5.5.4
futures==3.2.0
gunicorn==19.9.0
h5py==2.8.0
hyperlink==18.0.0
idna==2.7
incremental==17.5.0
jmespath==0.9.3
ndg-httpsclient==0.3.2
numexpr==2.6.8
numpy==1.15.1
packaging==16.8
pandas==0.23.4
ply==3.4
pyasn1==0.1.7
pyasn1-modules==0.0.5
pycparser==2.18
pycrypto==2.6.1
[...]

PR #806, opened 2018-09-01
import setUpRandomResolver only for new Twisted

deniszh:
`twisted.internet._resolver` was introduced only in Twitter 17.1 (released on 11 Feb 2017).
I do not want to raise required Twisted version for a minor carbon update, so, this PR just making `setUpRandomResolver` import optional, because it's using only if DESTINATIONS_POOL_REPLICAS enabled (not by default).
Fixing issue #805.

/cc @iksaif @ybizeul

### DIFF
--- a/lib/carbon/client.py
+++ b/lib/carbon/client.py
@@ -12,7 +12,6 @@ from carbon.conf import settings
 from carbon.util import pickle
 from carbon.util import PluginRegistrar
 from carbon.util import enableTcpKeepAlive
-from carbon.resolver import setUpRandomResolver
 from carbon import instrumentation, log, pipeline, state
 
 try:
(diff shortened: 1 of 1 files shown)


deniszh:
OTOH it can be simplified to 
in lib/carbon/client.py
```
if settings.DESTINATION_POOL_REPLICAS:
  from carbon.resolver import setUpRandomResolver
else:
    setUpRandomResolver = None
```
and later
```
if setUpRandomResolver:
          setUpRandomResolver(reactor)
```
But in case of old Twisted it will produce error as in #805 and you'll need process exception there. I'm preferring explicit way instead.

lib/carbon/client.py
```diff
@@ -29,6 +28,17 @@
...
+  import twisted as _twisted
+  twisted_version = tuple(map(int, getattr(_twisted, '__version__', '0').split('.')))
+  if twisted_version >= (17, 1):
+    from carbon.resolver import setUpRandomResolver
+  else:
+    setUpRandomResolver = None
+  del _twisted
+except ImportError:
+  setUpRandomResolver = None
+
```
piotr1212: Wouldn't the following be enough?
```
try:
    from carbon.resolver import setUpRandomResolver
except ImportError:
    setUpRandomResolver = None
```

deniszh: Right, I overcomplicate it. Thanks!

deniszh:
Can I get more reviews for this, pretty please? I want to release 1.1.4 tonight. :)
/cc @piotr1212 @DanCech @iksaif 
Thanks a lot!

iksaif:
Sorry I'm out for a week, I'll trust you on that. Thanks !

Le lun. 3 sept. 2018 à 15:39, Denis Zhdanov <notifications@github.com> a
écrit :

PR #807, opened 2018-09-01
[1.1.x] Stop accept() when we reach MAX_RECEIVER_CONNECTIONS | Add DESTINATIONS_POOL_REPLICAS | client: deal correctly with cases where destinations are down | client: fix USE_RATIO_RESET with DESTINATION_POO

deniszh:
Backports the following commits to 1.1.x:
 - Stop accept() when we reach MAX_RECEIVER_CONNECTIONS (#800)
 - Add DESTINATIONS_POOL_REPLICAS (#801)
 - client: deal correctly with cases where destinations are down (#801)
 - client: fix USE_RATIO_RESET with DESTINATION_POOL_REPLICAS (#801)
 - Add a real random resolver for DESTINATIONS_POOL_REPLICAS (#801)
 - client: log errors in all deferred (#801)
 - routers: do not fail if there are 0 nodes (#801)
 - resolver: lint (#801)
 - cabon.conf.example: fix typo (#801)
 - Open stanza and 3.x compatibility (#802)
 - routers: return instead of StopIteration() (#801)


deniszh:
Backport of latest PRs to 1.1.x

PR #812, opened 2018-09-12
WHISPER_FALLOCATE_CREATE documentation improved

haron:
Documenting silent conflict between WHISPER_SPARSE_CREATE and WHISPER_FALLOCATE_CREATE carbon-cache configuration variables: when both are set to True, WHISPER_SPARSE_CREATE is effectively disabled.


deniszh:
Thanks!

PR #817, opened 2018-10-18
[DISCUSS] release threads between writer runs

DanCech:
I have received a report of the writer thread stopping, which results in the metric cache growing and nothing being written to disk.

I'm not sure what the underlying cause is, though it's possible that accessing `reactor.running` from the background thread is part of the issue.  I also noted that when the thread calls `time.sleep()` between runs it is blocking for no good reason.

This PR modified the `writeForever` and `writeTagsForever` functions to run as regular twisted async functions, using `reactor.callLater()` to schedule their next run, and `threads.deferToThread` to call `writeCachedDataPoints` and `writeTags` in threads from the twisted threadpool.  This avoids the need to access `reactor` from those threads, and leaves scheduling in the main thread.

The only behavior change this should introduce is that I set the wait time after draining the queue to 1s for `writeTagsForever` to match `writeForever`, though I do wonder whether it would make more sense to shorten that.

I have not yet done any load testing on this, and would appreciate any feedback.

### DIFF
--- a/lib/carbon/writer.py
+++ b/lib/carbon/writer.py
@@ -186,7 +186,7 @@ def writeCachedDataPoints():
 def writeForever():
   def handleOk(result):
     # Avoid churning CPU when there are no series in the queue
-    reactor.callLater(1, writeForever)
+    reactor.callLater(0.2, writeForever)
 
   def handleErr(err):
     log.msg('Error writing data points: %s' % err)
@@ -207,7 +207,7 @@ def writeTags():
 def writeTagsForever():
   def handleOk(result):
     # Avoid churning CPU when there are no series in the queue
-    reactor.callLater(1, writeTagsForever)
+    reactor.callLater(0.2, writeTagsForever)
 
   def handleErr(err):
     log.msg('Error writing tags: %s' % err)
--- a/lib/carbon/writer.py
+++ b/lib/carbon/writer.py
@@ -22,7 +22,7 @@ from carbon.conf import settings
 from carbon import log, instrumentation
 from carbon.util import TokenBucket
 
-from twisted.internet import reactor
+from twisted.internet import reactor, threads
 from twisted.internet.task import LoopingCall
 from twisted.application.service import Service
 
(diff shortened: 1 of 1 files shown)


deniszh:
Well, it should work. Not sure how we can test that properly, though. I'm working on cluster implementation on docker compose, maybe I can add it there, and test it using haggar...

piotr1212:
Seems like a good thing to do. I was already wondering about the 1 second sleep before, seems like a bit long.

deniszh:
Yes, still working on that. Docker image is configurable now, so, I'm gonna create docker compose setup with haggar and test how it works.

deniszh:
Ok, benchmarking is hard. :)
I did a small test of this patch vs 1.1.4 using this setup - https://github.com/graphite-project/graphite-test-docker (MBP 2017 / 4CPU / 4G in Docker, 10 clients x 1000 metrics every 10 seconds, run for 1 hour)
1.1.4:
screenshot 2018-11-04 at 10 55 33
screenshot 2018-11-04 at 10 55 45
screenshot 2018-11-04 at 10 55 52
This patch:
screenshot 2018-11-04 at 15 05 37
screenshot 2018-11-04 at 15 05 48
screenshot 2018-11-04 at 15 05 57
I.e. on the first sight it works fine. You can use the same repo to run own tests.

deniszh:
OK, what we're gonna do with that? go / no-go? @DanCech @piotr1212 @iksaif ?

piotr1212:
Your graphs look different between runs but your conclusion is that it looks good? cpu, load, number of metrics in cache, memory used are all higher. 

I don't expect this to have any influence on performance but don't have time to test this right now.

deniszh:
The second graph is after 3-4 hours and first - after the first hour. As I said benchmarking is hard and I don't even try to do so, only provide (quite weak) proof that this PR is not horribly blown in your face.

DanCech:
Heh, I'd forgotten about this one.  Seems like it would still be a good idea

PR #826, opened 2018-11-16
Make max pickle string length configurable

semyonslepov:
By default, max length for a pickle string is 1mb and it's not configurable. I have recently hit this limit on our setup where statsd collects data from many other senders and sends it to Graphite.
This change allows to configure the max length.


piotr1212:
Seems like a good addition, thanks.

I think it would be better to have the setting name start with PICKLE_RECEIVER_ as it only applies to the pickle receiver (and not the carbonlink) and would me more in line with the existing options. I suggest PICKLE_RECEIVER_MAX_LENGTH.

And it would be a good thing to add it to the example.conf with a short description, somewhere here:
https://github.com/graphite-project/carbon/blob/master/conf/carbon.conf.example#L107

Issue #822, opened 2018-11-06
carbon-cache errors with "dictionary changed size during iteration" on carbon 1.1.4


standaloneSA:
I haven't completely narrowed down what is changing, but after running properly for some time, carbon-cache is failing with an unhandled error: 

    01/11/2018 10:52:31 :: Unhandled Error
    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/dist-packages/twisted/python/threadpool.py", line 250, in inContext
        result = inContext.theWork()
      File "/usr/local/lib/python3.5/dist-packages/twisted/python/threadpool.py", line 266, in <lambda>
        inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
      File "/usr/local/lib/python3.5/dist-packages/twisted/python/context.py", line 122, in callWithContext
        return self.currentContext().callWithContext(ctx, func, *args, **kw)
      File "/usr/local/lib/python3.5/dist-packages/twisted/python/context.py", line 85, in callWithContext
        return func(*args,**kw)
    --- <exception caught here> ---
      File "/opt/graphite/lib/carbon/writer.py", line 189, in writeForever
        writeCachedDataPoints()
      File "/opt/graphite/lib/carbon/writer.py", line 98, in writeCachedDataPoints
        (metric, datapoints) = cache.drain_metric()
      File "/opt/graphite/lib/carbon/cache.py", line 187, in drain_metric
        metric = self.strategy.choose_item()
      File "/opt/graphite/lib/carbon/cache.py", line 116, in choose_item
        return next(self.queue)
      File "/opt/graphite/lib/carbon/cache.py", line 104, in _generate_queue
        metric_counts = sorted(self.cache.counts, key=lambda x: x[1])
      File "/opt/graphite/lib/carbon/cache.py", line 161, in counts
        return [(metric, len(datapoints)) for (metric, datapoints) in self.items()]
      File "/opt/graphite/lib/carbon/cache.py", line 161, in <listcomp>
        return [(metric, len(datapoints)) for (metric, datapoints) in self.items()]
    builtins.RuntimeError: dictionary changed size during iteration

From what I can tell, this may be an incompatability with Python 3 ( per https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error ). 

The machine is Ubuntu 16.04, with carbon 1.1.4 installed from pip. 

If I can give any other information on the machine or environment, please let me know. Thanks! 

--Matt

[...]

Issue #815, opened 2018-10-07
Recurrent uncaught exception in choose_metric


benjolitz:
Found this spamming my syslog:

```
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] Unhandled Error
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	Traceback (most recent call last):
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/usr/local/lib/python3.5/site-packages/twisted/python/threadpool.py", line 250, in inContext
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    result = inContext.theWork()
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/usr/local/lib/python3.5/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/usr/local/lib/python3.5/site-packages/twisted/python/context.py", line 122, in callWithContext
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    return self.currentContext().callWithContext(ctx, func, *args, **kw)
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/usr/local/lib/python3.5/site-packages/twisted/python/context.py", line 85, in callWithContext
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    return func(*args,**kw)
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	--- <exception caught here> ---
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/opt/graphite/lib/carbon/writer.py", line 188, in writeForever
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    writeCachedDataPoints()
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	  File "/opt/graphite/lib/carbon/writer.py", line 97, in writeCachedDataPoints
Oct  6 11:40:24 nyx carbon-cache-a[1193]: [console] 	    (metric, datapoints) = cache.drain_metric()
[...]

PR #829, opened 2018-12-07
Fix carbon cache locking

piotr1212:
Add lock when calling choose_item(), this seemed to cause the issues in #822 #815.
Remove previously added locks to counts and watermarks as they deadlocked with
the new lock.
Improve locking in store(), it would lock _after_ checking if the timestamp is in
the cache. The cache could have been altered between checking and locking.

### DIFF
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -158,16 +158,14 @@ class _MetricCache(defaultdict):
 
   @property
   def counts(self):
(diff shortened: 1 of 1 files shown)


deniszh:
I think that should be merged, right, @piotr1212 ?

piotr1212:
Would be nice to have some more regression testing.

PR #835, opened 2019-01-08
Remove /opt/graphite prefix and use setuptools

piotr1212:
This switches from distutils to setuptools and removes the default `/opt/graphite` prefix.

I think this should go in a major or at least a minor release (no patch release).

Reasons for removing `/opt/graphite` prefix:

- Does not work with setuptools
- Confuses new users/contributors
- Is not common among Python applications
- It is unclear how to disable it
- Causes issues with PYTHONPATH (just google "No module named 'graphite'")
- The code that accomplishes this in setup.py is ugly (rewriting setup.cfg)
- Causes unexpected issues, e.g. https://github.com/graphite-project/carbonate/issues/101
- There are better ways to install in other locations.
- Makes installation of Graphite "harder", have to account for it in your Ansible playbooks/Chef recipes/Puppet Modules.
- Requires `root` privileges on many hosts which don't have a /opt partition
- Pisses me off every time I forget to set the `GRAPHITE_NO_PREFIX` environment variable.

I don't see any reasons for keeping it other than:

- This has been the default for 10 years

If you need to install Graphite in an other location you should use:

- virtualenv/conda (recommended)
- `--user` switch
- `--user` switch and a customized `PYTHONUSERBASE`
- maybe even Docker

[more on custom locations](https://setuptools.readthedocs.io/en/latest/easy_install.html#custom-installation-locations)

Switch to setuptools gets rid of the following warnings:
`UserWarning: Unknown distribution option: 'install_requires'`
`UserWarning: Unknown distribution option: 'long_description_content_type'`

PPA discourages the use of distutils and suggests setuptools. Setuptools has some nice features we can use in future.

### DIFF
--- a/bin/carbon-client.py
+++ b/bin/carbon-client.py
@@ -29,7 +29,6 @@ from carbon.routers import ConsistentHashingRouter, RelayRulesRouter  # noqa
 from carbon.client import CarbonClientManager  # noqa
 from carbon import log, events  # noqa
 
-CONF_DIR = join(sys.prefix, 'conf')
 default_relayrules = join('/opt/graphite', 'relay-rules.conf')
 
 option_parser = OptionParser(usage="%prog [options] <host:port:instance> <host:port:instance> ...")
--- a/bin/carbon-client.py
+++ b/bin/carbon-client.py
@@ -14,19 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 
 import sys
-from os.path import dirname, join, abspath, exists
+from os.path import join, exists
 from optparse import OptionParser
 
-# Figure out where we're installed
-BIN_DIR = dirname(abspath(__file__))
-ROOT_DIR = dirname(BIN_DIR)
-CONF_DIR = join(ROOT_DIR, 'conf')
+CONF_DIR = join(sys.prefix, 'conf')
 default_relayrules = join(CONF_DIR, 'relay-rules.conf')
-
-# Make sure that carbon's 'lib' dir is in the $PYTHONPATH if we're running from
-# source.
-LIB_DIR = join(ROOT_DIR, 'lib')
-sys.path.insert(0, LIB_DIR)
 
 try:
   from twisted.internet import epollreactor
--- a/bin/carbon-client.py
+++ b/bin/carbon-client.py
@@ -17,9 +17,6 @@ import sys
 from os.path import join, exists
 from optparse import OptionParser
 
-CONF_DIR = join(sys.prefix, 'conf')
-default_relayrules = join(CONF_DIR, 'relay-rules.conf')
-
 try:
   from twisted.internet import epollreactor
   epollreactor.install()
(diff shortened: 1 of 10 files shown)


piotr1212:
We should probably also get rid of code snippets like these https://github.com/graphite-project/carbon/blob/master/bin/carbon-cache.py#L19-L26

DanCech:
I like this idea, the biggest hurdle for new users imho is how complicated it is to get set up so anything we can do to streamline things is great! We do need to make sure we handle upgrades properly and that the installation/upgrade docs are all clear and accurate.

deniszh:
Yep, that's a good question. We should provide nice doc to upgrade graphite which still installed in `/opt/graphite` to new setup.

piotr1212:
Good points, I can take care of the docs, but that can take a while, I'm a
bit busy the next two weeks.

I'll also have to check all example config files for paths.

ploxiln:
(copying over a comment from https://github.com/graphite-project/graphite-web/pull/2409#issuecomment-465734061)

I had another thought about storage dirs ... I think the original idea behind using /opt/graphite is those storage and log dirs, which would be awkward in the python site-packages directory. Maybe the thing to do is divorce them from the graphite application root, and default to /opt/graphite/storage and /opt/graphite/log regardless of the install prefix? And not mention them in setup.py at all? I suppose the downside is they would not be created by install. Just an idea - I suppose I always customize these dirs anyway.

deniszh:
That could be good solution, @ploxiln

piotr1212:
Yup, I don't think how I did it is the right way. Maybe instead of /opt/graphite use something more FHS compliant like /var/lib/graphite and /var/log/graphite.

setup.py
```diff
@@ -1,115 +1,60 @@
...
 install_files = storage_dirs + conf_files
 
 # Let's include redhat init scripts, despite build platform
 # but won't put them in /etc/init.d/ automatically anymore
-init_scripts = [ ('examples/init.d', ['distro/redhat/init.d/carbon-cache',
-                                      'distro/redhat/init.d/carbon-relay',
-                                      'distro/redhat/init.d/carbon-aggregator']) ]
+init_scripts = [('examples/init.d', ['distro/redhat/init.d/carbon-cache',
+                                     'distro/redhat/init.d/carbon-relay',
+                                     'distro/redhat/init.d/carbon-aggregator'])]
```
ploxiln, line 19: I suggest taking the opportunity to remove these init script examples from the install.

piotr1212: good point

piotr1212:
Reverted graphite_root to /opt/graphite. This change is about not installing the python program code in /opt/graphite, can still use /opt/graphite for config and storage.

bin/carbon-client.py
```diff
@@ -40,6 +29,8 @@
 from carbon.client import CarbonClientManager  # noqa
 from carbon import log, events  # noqa
 
+CONF_DIR = join(sys.prefix, 'conf')
```
ploxiln, line 32: another `sys.prefix` that should be `/opt/graphite`?

ploxiln: (actually, this `CONF_DIR` variable is now unused)

deniszh:
I think it's important, will pin it.

piotr1212:
I really want to have this done but don't see that I will have the time anytime soon.

deniszh:
Hi @piotr1212 ,

No worries, I assigned you because it's your PR, it's not an obligation for you to finish that.

vincele:
This look like a really good cleanup, what is missing to allow progress on the subject ?

The hardcoded `/opt/graphite` is a bit sad though, I'd prefer more classical `/var/{lib,log}/graphite` defaults.

deniszh:
Unfortunately, that 's incompatible change. Probably, still can be done, together with major release.

PR #859, opened 2019-07-08
Add testing for Python 3.8

piotr1212:


piotr1212:
Tests fail due to:
- issue in Python's C-API, which is fixed in beta2
- issue in Pyhash, fixed in master https://github.com/flier/pyfasthash/issues/35

Let's wait a week or so before merging to have green test results.

PR #869, opened 2019-12-10
Add s390x support for travis

sangitanalkar:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,33 @@ matrix:
     - python: "3.8-dev"
       env:
         - TOXENV=lint
+    - env:
+        - TOXENV=py27
+      arch: s390x
+    - env:
+        - TOXENV=lint
+      arch: s390x
+    - python: 3.5
+      env:
+        - TOXENV=py35
+      arch: s390x
+    - python: 3.6
+      env:
+        - TOXENV=py36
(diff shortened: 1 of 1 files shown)


piotr1212:
Not against, but is there actually anyone who will fix s390 issues if they would occur?

Is it needed to duplicate all entries? Wouldn't just adding an "arch:" entry work like described [here](https://docs.travis-ci.com/user/multi-cpu-architectures/#testing-on-multiple-cpu-architectures).

sangitanalkar:
Hi,

Yes, we already build and maintain carbon on s390x. Our build instructions are published [here ](https://github.com/linux-on-ibm-z/docs/wiki/Building-Graphite-Carbon).

Regarding "Is it needed to duplicate all entries? ", if we just add "arch: " entry, only 3 travis builds are triggered for s390x for the 3 environments mentioned at " https://github.com/graphite-project/carbon/blob/master/.travis.yml#L34 "  and the ones in the matrix are not triggered. Hence added it separately.

sangitanalkar:
Hi, any updates on the PR if the changes can be merged?

PR #872, opened 2020-01-13
Fix #871: Adjust aggregator-rules input_pattern match greediness

### LINKED ISSUE #871: [BUG] carbon-aggregator input regexp match is greedy
**Describe the bug**
When using aggregator to aggregate metrics from nodes with a name such as "mycluster-location1" to "mycluster-location350" using a regexp with "mycluster-\<location\>\d+" the rule regexp breaks it out to summaries for "location", "location1", "location2" ... up to "location35" instead of "location".

**To Reproduce**
Configure aggregation-rule of:

aggregate.\<location\>.app.requests (60) = sum collectd.mycluster-\<location\>\d+.app.requests

Feed data with collectd.mycluster-helsinki150.app.requests, observe aggregated data point come out for aggregate.helsinki15.app.requests.

**Expected behavior**
Should not include the numbers matched by \d+ in the aggregate identiifer.

The issue appears to be in lib/carbon/aggregator/rules.py, regex_part setup, it should use the '?' non-greedy modifier suffix to make the match not greedy and let the user-supplied part on the right side be greedy. I have a tested patch, I will make a pull request shortly.

hessu:
... to support numeric matching after captured field. Fixes bug #871 .

Before, an input regexp of "collectd.mycluster-\<location\>\d+.app.requests" in aggregation-rules.conf, with a hostname of "mycluster-paris105" would capture "paris10" as the location, instead of paris. With a large cluster this would create a whole lot of aggregates, one per 10 nodes.

This patch makes the \<location\> match not greedy, letting the user-supplied \d+ side to be greedy and eat the 105 away fully, leaving "paris" as the location.

### DIFF
--- a/lib/carbon/aggregator/rules.py
+++ b/lib/carbon/aggregator/rules.py
@@ -128,7 +128,7 @@ class AggregationRule(object):
         pre = input_part[:i]
         post = input_part[j + 2:]
         field_name = input_part[i + 2:j]
-        regex_part = '%s(?P<%s>.+)%s' % (pre, field_name, post)
+        regex_part = '%s(?P<%s>.+?)%s' % (pre, field_name, post)
 
       else:
         i = input_part.find('<')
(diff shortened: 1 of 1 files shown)


piotr1212:
@deniszh It seems that pyhash does not support py27 anymore which results in failing travis builds. We should drop py27-pyhash or pin pyhash to some older version.

deniszh:
@piotr1212 : I'm voting for drop, but pinning also OK for me.

@hessu : thanks for fix, LGTM!

PR #874, opened 2020-01-19
Another test fix try

deniszh:


deniszh:
Ok, by default Travis use system `pip` and `virtualenv`. I tried to upgrade it before - it works fine now. 
Should we do that for all envs? Or keep it only for 2.7?
@piotr1212 @DanCech @ploxlin ?

deniszh:
Looking for [Graphite-web results](https://github.com/graphite-project/graphite-web/pull/2530) I think it's a good idea to test on latest pip and virtualenv.

ploxiln:
I'd lean towards just upgrading pip/virtualenv for python2.7 - following the other patterns, something like

```
  - if [[ '$TOXENV' == py2* ]]; then pip install --upgrade pip virtualenv; fi
```

(I suspect the `bash -c "..."` wrapper isn't necessary, but that's a separate thing that could be simplified later I guess.)

But anyway, no significant objection to doing it your way, that's also fine.

deniszh:
@ploxiln:  Maybe you're right. Let's stick to 2.7 for now.

ploxiln:
looks good, thanks :)
👍

deniszh:
Will merge that for now and trigger tests in all non-merged PRs.

PR #877, opened 2020-02-02
Adding Pipfile

deniszh:
And excluding Pipfile.lock from repo


deniszh:
A-ha, tests for s390 became broken, that's pity

PR #879, opened 2020-02-04
Bucketmax write strategy

piotr1212:
Bucketmax writes the metric with the most datapoints just as max does but uses
a different approach to determine which metric has the most datapoints.

At each write operation 'max' loops once over the list of metrics and picks the metric
with most datapoints.

'bucketmax' stores all metrics in a list where each index represents the number of
datapoints in the specific metrics. For example `buckets[5]` contains a list of all
metrics which have 6 datapoints. At each write operation 'bucketmax' selects the
first metric from the largest bucket.

The time complexity of 'max' is linear, 'bucketmax' is near constant time but runs
on every ingestion where 'max' only runs on every write. There is also some overhead
in resizing the buckets list. I've done some limited testing and 'bucketmax'' CPU
usage was almost half of that of 'max'. The constraint of bucketmax is that the number
of datapoints per metrics must be relatively low compared to the total number of
unique metrics (which I think is the most common use-case in Graphite usage). I don't
know where the tipping point is but if one has for example 10 metrics with each 100000
datapoints I would assume 'bucketmax' would perform worse than 'max'.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -165,7 +165,7 @@ LOG_CACHE_QUEUE_SORTS = False
 # disk in that order.
 #
 # timesorted - All metrics in the list will be looked at and sorted according
-# to the timestamp of there datapoints. The metric that were the least recently
+# to the timestamp of their datapoints. The metric that were the least recently
 # written will be written first. This is an hybrid strategy between max and
 # sorted which is particularly adapted to sets of metrics with non-uniform
 # resolutions.
@@ -176,6 +176,12 @@ LOG_CACHE_QUEUE_SORTS = False
 # updated metrics may only ever be persisted to disk at daemon shutdown if
 # there are a large number of metrics which receive very frequent updates OR if
 # disk i/o is very slow.
+#
+# bucketmax (experimental) - As 'max' but uses a different algorithm to
+# determine the metric with the most datapoints.
+# Should perform better than 'max' on most loads, could perform worse if the
+# number of metrics is very small while the number of datapoints per metric is
+# very high.
 #
 # naive - Metrics will be flushed from the cache to disk in an unordered
 # fashion. This strategy may be desirable in situations where the storage for
--- a/lib/carbon/cache.py
+++ b/lib/carbon/cache.py
@@ -57,6 +57,9 @@ class DrainStrategy(object):
   def choose_item(self):
     raise NotImplementedError()
 
+  def store(self, metric):
+    pass
+
 
 class NaiveStrategy(DrainStrategy):
   """Pop points in an unordered fashion."""
@@ -146,6 +149,41 @@ class TimeSortedStrategy(DrainStrategy):
     return next(self.queue)
 
 
+class BucketMaxStrategy(DrainStrategy):
+    """
+    Same as 'max' strategy but sorts on insertion into buckets instead of at
+    pop().
+    """
+    def __init__(self, cache):
+        self.buckets = list()
+        super(BucketMaxStrategy, self).__init__(cache)
+
+    def choose_item(self):
+        try:
+            # Largest buckets are empty, remove them.
+            while len(self.buckets[-1]) == 0:
+                self.buckets.pop()
+            # return the metric with the most datapoints. If there is
+            # more than one metrics which has the most datapoints the
+            # first seen is returned.
+            return self.buckets[-1].pop(0)
+        except (KeyError, IndexError):  # buckets are empty
+            return None
+
+    def store(self, metric):
+        nr_points = len(self.cache[metric])
+
+        # No bucket of this size exists, create it
+        while nr_points > len(self.buckets):
+            self.buckets.append(list())
+
+        # Remove existing metrics from its bucket
+        if nr_points > 1:
+            self.buckets[nr_points - 2].remove(metric)
+
+        self.buckets[nr_points - 1].append(metric)
+
+
 class _MetricCache(defaultdict):
   """A Singleton dictionary of metric names and lists of their datapoints"""
   def __init__(self, strategy=None):
@@ -217,6 +255,8 @@ class _MetricCache(defaultdict):
         else:
           self.size += 1
           self[metric][timestamp] = value
+          if self.strategy:
+            self.strategy.store(metric)
       else:
         # Updating a duplicate does not increase the cache size
         self[metric][timestamp] = value
@@ -232,7 +272,7 @@ def MetricCache():
 
   # Initialize a singleton cache instance
   # TODO: use plugins.
-  write_strategy = None
+  write_strategy = DrainStrategy
   if settings.CACHE_WRITE_STRATEGY == 'naive':
     write_strategy = NaiveStrategy
   if settings.CACHE_WRITE_STRATEGY == 'max':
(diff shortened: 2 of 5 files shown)


conf/carbon.conf.example
```diff
@@ -177,6 +177,9 @@ LOG_CACHE_QUEUE_SORTS = False
 # there are a large number of metrics which receive very frequent updates OR if
 # disk i/o is very slow.
 #
+# bucketmax (experimental) - As 'max' but uses a different algorithm to determine
+# the metric with the most datapoints. Should perform better than 'max'.
+#
```
deniszh: Maybe you should mention " number of datapoints per metrics vs total number of
unique metrics" from PR description here too.

piotr1212: I guess so. Thing is that that statement is an assumption. This whole thing needs to be better tested to see how it behaves in real life and if my assumption is correct.

deniszh: Sure, my idea was to extend that a bit for users, e.g. "Should perform better than 'max' if number of datapoints per metric is relatively low compared to the total number of
unique metrics, which should be common use-case. See https://github.com/graphite-project/carbon/pull/879 for details."
But that's optional ofc.

deniszh:
Great idea, I like that, @piotr1212 !
For me it looks like much proper `max` stratege, I mean, for typical use-case.

DanCech:
I don't see where the metric is removed from the associated bucket when it is flushed?

I'm also concerned about the impact on memory usage of maintaining the buckets, since you're going to have a second copy of all the metric names in the bucket data structure.

piotr1212:
> I don't see where the metric is removed from the associated bucket when it is flushed?

It is being popped here: https://github.com/graphite-project/carbon/pull/879/files#diff-b95553eddf4a9a25750a9a1d17f2be0cR161

> I'm also concerned about the impact on memory usage of maintaining the buckets, since you're going to have a second copy of all the metric names in the bucket data structure.

Good point, i'll see if I can store pointers instead of the full string somehow.

piotr1212:
It seems that Python does not copy the metric name but just stores a reference to the string object.

See below for a small simulation, the refcount to the metric name increases when I add it to the buckets. The size of the buckets is smaller than the actual size of the string.


```
In [62]: a_metric_name = 'a very long string representing a metric name in graphite, foobar, some more text, even more and more and more and more'            

In [63]: sys.getsizeof(a_metric_name)  
Out[63]: 168

In [64]: sys.getrefcount(a_metric_name)
Out[64]: 2

In [65]: a_metric_cache = dict()       

In [66]: sys.getsizeof(a_metric_cache) 
Out[66]: 248

In [67]: a_metric_cache[a_metric_name] = [1, 2, 3, 4, 5]                       

In [68]: sys.getsizeof(a_metric_cache) 
Out[68]: 248

In [69]: sys.getrefcount(a_metric_name)
Out[69]: 3

In [70]: buckets = list()              

In [71]: sys.getsizeof(buckets)        
Out[71]: 72

In [72]: buckets.append(a_metric_name) 

In [73]: sys.getsizeof(buckets)        
Out[73]: 104

In [74]: sys.getrefcount(a_metric_name)
Out[74]: 4
```

DanCech:
That's good to know.  The only other comment I'd have would be around how we're now mingling handling between the strategy and the cache, and that the cache has strategy-specific code in it.

One thought was that the buckets could be defined within the strategy and a method on the strategy called each time we add a new point so that it could do any housekeeping.  That would remove the strategy-specific code from the cache.

conf/carbon.conf.example
piotr1212: added some more info

piotr1212:
> That's good to know. The only other comment I'd have would be around how we're now mingling handling between the strategy and the cache, and that the cache has strategy-specific code in it.
> 
> One thought was that the buckets could be defined within the strategy and a method on the strategy called each time we add a new point so that it could do any housekeeping. That would remove the strategy-specific code from the cache.



Thank you for the suggestion. Updated, code looks much cleaner now.


> That's good to know. The only other comment I'd have would be around how we're now mingling handling between the strategy and the cache, and that the cache has strategy-specific code in it.
> 
> One thought was that the buckets could be defined within the strategy and a method on the strategy called each time we add a new point so that it could do any housekeeping. That would remove the strategy-specific code from the cache.

Thank you for the suggestion. Updated it, code looks much cleaner now.

lib/carbon/cache.py
```diff
@@ -146,6 +149,41 @@ def choose_item(self):
...
+            return None
+
+    def store(self, metric):
+        nr_points = len(self.cache[metric])
+
+        # No bucket of this size exists, create it
+        while nr_points > len(self.buckets):
+            self.buckets.append(list())
+
+        # Remove existing metrics from it's bucket
```
DanCech:
```suggestion
        # Remove existing metrics from its bucket
```

piotr1212:
squashed the commit

deniszh:
Cool! will try to run in on https://github.com/graphite-project/graphite-test-docker this weekend

deniszh:
Ok, I made graph below on my MBP 2017 (docker had 4 cores/6GB RAM) using https://github.com/graphite-project/graphite-test-docker
haggar generated -agents=10 -metrics=10000 = 100K metrics
```
MAX_CACHE_SIZE = Inf
MAX_UPDATES_PER_SECOND = 5000
MAX_CREATES_PER_MINUTE = 50
```

`CACHE_WRITE_STRATEGY = max`
Screenshot 2020-02-08 at 18 49 27
Screenshot 2020-02-08 at 18 49 40
Screenshot 2020-02-08 at 18 50 04

`CACHE_WRITE_STRATEGY = bucketmax`
Screenshot 2020-02-08 at 23 35 57
Screenshot 2020-02-08 at 23 36 15
Screenshot 2020-02-08 at 23 36 33

piotr1212:
But CPU is nearly identical, memory usage is higher. That is quite different from what I measured. I'll do some more testing

piotr1212:
50k metrics every 10s, metrics were already existing.

carbon.conf
``` 
MAX_CACHE_SIZE = inf
MAX_UPDATES_PER_SECOND = 500
MAX_CREATES_PER_MINUTE = 1000000
```

load generator is  https://github.com/Civil/graphite_perf_test_go
```
~/go/bin/graphite_perf_test_go -host 127.0.0.1:2003 -connec
tions 50 -interval 10                                                          
2020/02/09 16:45:33 Starting...                                                
2020/02/09 16:45:33 ===New iteration===                  
2020/02/09 16:45:33 Load    :  50 x 1000 = 50000  metrics
2020/02/09 16:45:34 Spent   :  0.043315 seconds          
2020/02/09 16:45:34 Speed   :  1154334.5 metrics/second  
2020/02/09 16:45:34 Sleeping:  9.956684 seconds          
```


first 30m is bucketmax, then 15m max, then bucketmax again. You see a significant lower CPU usage with bucketmax.

b_max

deniszh:
> But CPU is nearly identical, memory usage is higher. That is quite different from what I measured. I'll do some more testing

Well, memory is the same, I just waited more time. But you're right, some owerflowing test has much more sense here.

piotr1212:
Ok, merging this.

Dieterbe:
Is there an explanation for the significantly reduced pointsPerUpdate? I thought the behavior would be equivalent.

piotr1212:
With the 'max' strategy CPU goes to 100%. So I assume it does less write
operations which results in larger queues so more points per update. I
probably should have added the 'write_operations' to the graph and do a
test where it does not overload the CPU.

Dieterbe:
Ah that makes sense and explains why pointsPerUpdate grows linearly.  This also means the memory usage in non overload scenario is likely about the same as judging from the start of the 'max's run

PR #881, opened 2020-02-07
Trying to fix tests

deniszh:
Not sure why we mocking 'carbon.state.instrumentation' here.
It recently start to give following error:
```python
Traceback (most recent call last):
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/twisted/trial/runner.py", line 531, in loadPackage
    module = modinfo.load()
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/twisted/python/modules.py", line 392, in load
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/twisted/python/reflect.py", line 308, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/twisted/python/reflect.py", line 247, in _importAndCheckStack
    return __import__(importName)
  File "/Users/dzhdanov/git/carbon/lib/carbon/tests/test_client.py", line 91, in <module>
    @patch('carbon.state.instrumentation', Mock(spec=instrumentation))
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/mock/mock.py", line 1084, in __init__
    _spec_state, _new_name, _new_parent, **kwargs
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/mock/mock.py", line 439, in __init__
    self._mock_add_spec(spec, spec_set, _spec_as_instance, _eat_self)
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/mock/mock.py", line 494, in _mock_add_spec
    if iscoroutinefunction(getattr(spec, attr, None)):
  File "/Users/dzhdanov/git/carbon/.tox/py37/lib/python3.7/site-packages/mock/backports.py", line 34, in iscoroutinefunction
    getattr(obj, '_is_coroutine', None) is _is_coroutine
builtins.KeyError: '_is_coroutine'
```
Commenting out works fine.
Did someone maybe know why that happening and how we can fix it properly?
@piotr1212 @DanCech ?

PR #880, opened 2020-02-07
Fixing tests for S390

deniszh:
Testing fix from https://github.com/graphite-project/carbon/issues/878#issuecomment-583374163

PR #883, opened 2020-02-21
[1.1.x] s390x support for travis (#869) | Fix #871: Adjust aggregator-rules input_pattern match greediness to support numeric matching after captured field (#872) | sanitize names when using them as tag value (#858) | sanitize names when us

### LINKED ISSUE #871: [BUG] carbon-aggregator input regexp match is greedy

deniszh:
Backports the following commits to 1.1.x:
 - s390x support for travis (#869)
 - Fix #871: Adjust aggregator-rules input_pattern match greediness to support numeric matching after captured field (#872)
 - sanitize names when using them as tag value (#858)
 - sanitize names when using them as tag value (#858)
 - sanitize names when using them as tag value (#858)
 - Another test fix try (#874)
 - simplify travis-ci config (#875)
 - Fixing tests for S390 (#880)
 - Trying to fix tests (#881)
 - Fix the manhole for Twisted > 16 and Python 3 (#882)


deniszh:
Preparing 1.1.7 release

piotr1212:
pr858 is added three times? You might want to check what went wrong there.

deniszh:
Same thing as with 2458 in https://github.com/graphite-project/graphite-web/pull/2545

It just inserting empty commits for some reason.

deniszh:
I mean, PR is good, despite duping

Issue #884, opened 2020-02-25
[BUG] Python3 and line protocol cause Data must not be unicode


pkruk:
**Describe the bug**
Hi I have a problem with official graphite-statsd docker container which use latest version of carbon.
When using line protocol, the twisted library have a problem with encoding:

```
  File "/opt/graphite/lib/python3.7/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/opt/graphite/lib/carbon/client.py", line 259, in sendQueued
    self.connectedProtocol.sendQueued()
  File "/opt/graphite/lib/carbon/client.py", line 135, in sendQueued
    self.sendDatapointsNow(self.factory.takeSomeFromQueue())
  File "/opt/graphite/lib/carbon/client.py", line 95, in sendDatapointsNow
    self._sendDatapointsNow(datapoints)
  File "/opt/graphite/lib/carbon/client.py", line 487, in _sendDatapointsNow
    self.sendLine("%s %s %d" % (metric, value, datapoint[0]))
  File "/opt/graphite/lib/python3.7/site-packages/twisted/protocols/basic.py", line 476, in sendLine
    return self.transport.writeSequence((line, self.delimiter))
  File "/opt/graphite/lib/python3.7/site-packages/twisted/internet/abstract.py", line 377, in writeSequence
    raise TypeError("Data must not be unicode")
builtins.TypeError: Data must not be unicode
```

To solve it I add to a code:

```
class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):                                                                                                                                             
                                                                                                                                                                                                                    
  def _sendDatapointsNow(self, datapoints):                                                                                                                                                                         
    for metric, datapoint in datapoints:                                                                                                                                                                            
      if isinstance(datapoint[1], float):                                                                                                                                                                           
        value = ("%.10f" % datapoint[1]).rstrip('0').rstrip('.')                                                                                                                                                    
      else:                                                                                                                                                                                                         
        value = "%d" % datapoint[1]                                                                                                                                                                                 
      to_send = "%s %s %d" % (metric,value, datapoint[0])                                                                                                                                                           
      self.sendLine(to_send.encode('utf-8'))   <--- My Change
```

Will you accept a PR with this change ?

pkruk:
I created a PR for that bug :)

deniszh:
Not sure if it's good solution and will not have some bad consequences...
@piotr1212 ?

[...]

PR #885, opened 2020-02-25
Fix missing encoding for line protocol

pkruk:
PR for issue: https://github.com/graphite-project/carbon/issues/884


piotr1212:
Thanks this should work but the tests need to be adjusted, see the travis results. You can ignore the deepsource results. There is also a trailing whitespace which breaks the lint check.

to run the tests locally just invoke `tox`.

deniszh:
I disable Deepsource, maybe will remove it completely.

pkruk:
@piotr1212 thanks! I will fix it!

piotr1212:
tests are still failing, you can replace the expected values with the actual ones (just add the b in front of the string to indicate it is a bytearray)

pkruk:
@piotr1212 Hey, I will fix it in a moment :) thanks!

PR #890, opened 2020-04-25
remove carbon-client.py, is broken for 4 years

piotr1212:
Trying to figure out what this does (never used before) looks like it is broken for a long time. Have not seen any complaints so maybe remove it instead of fixing?

Is anybody using this?


deniszh:
Looks like it's like carbon but takes metrics from stdin. Looks like broken for long time, agreed.

piotr1212:
Merging. If someone was actually using this in some old version then let me know.

PR #893, opened 2020-05-10
Spelling

jsoref:
Misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

It reported: https://github.com/jsoref/carbon/commit/42c4302633b1da410a9efac5d7072a102015ac7e#commitcomment-39064736

And it validated that the changes in this PR made it happy: https://github.com/jsoref/carbon/commit/8ba47d134691ae543c9a574adc4ff50897b94ac4

PR #894, opened 2020-05-10
Add skipcq: PTC-W0034

jsoref:
https://deepsource.io/gh/graphite-project/carbon/run/f7c9db6c-4530-4b82-8ae7-57850da0e1e0/python/PTC-W0034


deniszh:
@jsoref FYI, lint check is broken now:
```
lib/carbon/util.py:235:32: E261 at least two spaces before inline comment
lib/carbon/util.py:235:33: E262 inline comment should start with '# '
lib/carbon/util.py:257:32: E261 at least two spaces before inline comment
lib/carbon/util.py:257:33: E262 inline comment should start with '# '
```

jsoref:
Fwiw, I make no warranty about this being the right thing to do in general, it's just to silence the warnings as reported for #893

Issue #768, opened 2018-04-24
Latest change in storage-aggregation.conf.example breaks statsd counters


antonsoroko:
Previously for counters aggregationMethod was sum, now it is max
https://github.com/graphite-project/carbon/commit/e4953da4a37c07177fdccb1abda3679e017a83e8

This is good for use case described in https://answers.launchpad.net/graphite/+question/286462 but wrong for statsd counters which are reset every (statsd) flush interval.

Maybe it would be more reasonable to add a comment to the storage-aggregation.conf.example about which of `aggregationMethod` is suitable for different `counter` use cases.

What do you think?

DanCech:
That sounds like a good idea, and would clear up confusion about counters that reset every interval (statsd-style) vs counters that accumulate over time (snmp-style).

PR #899, opened 2020-08-16
Update storage-aggregation.conf.example for #768

antonsoroko:
see #768

Previously for counters aggregationMethod was sum, now it is max
https://github.com/graphite-project/carbon/commit/e4953da4a37c07177fdccb1abda3679e017a83e8

This is good for use case described in https://answers.launchpad.net/graphite/+question/286462 but wrong for statsd counters which are reset every (statsd) flush interval.


deniszh:
As I said 3 years ago - "max" is much more natural for counters.

PR #900, opened 2020-08-20
add config option to turn off logging of lost connections

tbenz9:
This patch adds a config option which can be added for the cache/relay/aggregator instances to **not** log lost connections anymore.
For this the config option `LOG_LISTENER_CONN_LOST` has to be set to `False`

Reason: Reduce the amount of logging from lost connections.

I tried to copy this MR as much as possible: https://github.com/graphite-project/carbon/pull/60


lib/carbon/conf.py
```diff
@@ -115,6 +115,7 @@
   TCP_KEEPINTVL=30,
   TCP_KEEPCNT=2,
   USE_RATIO_RESET=False,
+  LOG_LISTENER_CONN_LOST=True,
```
deniszh, line 118: I think we should keep compatibility with old versions. Could you please change default to False?

deniszh (changes requested):

lib/carbon/conf.py
tbenz9: I've updated this value to "False".

deniszh:
Thanks, but commented out values in config example still says "True" which can be misleading. Could you please change them too? Thanks!

tbenz9:
> Thanks, but commented out values in config example still says "True" which can be misleading. Could you please change them too? Thanks!

Done, thanks for your patience!

deniszh:
No worries, LGTM. Merging.

conf/carbon.conf.example
```diff
@@ -659,6 +667,10 @@ MAX_AGGREGATION_INTERVALS = 5
 # In order to turn off logging of successful connections for the line
 # receiver, set this to False
 # LOG_LISTENER_CONN_SUCCESS = True
+#
+# To turn off logging of lost connections for the line receiver set this to
+# False
```
ploxiln, line 672: Since the default is `False`, this should probably say "To turn on logging of lost connections, set this to True"

tbenz9: I'm happy to submit another MR with the change if you think it's important.  I personally think it's clear enough but I'm happy to do whatever the maintainers think is best.

PR #902, opened 2020-09-30
Bugfix/encoding in line protocol #884

pkruk:
According to what @brivadeneira mentioned here https://github.com/graphite-project/carbon/issues/884 I prepare a fix.

### DIFF
--- a/lib/carbon/client.py
+++ b/lib/carbon/client.py
@@ -479,6 +479,7 @@ class CarbonPickleClientFactory(CarbonClientFactory):
 class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
 
   def _sendDatapointsNow(self, datapoints):
+    self.delimiter = b'\r\n'
     for metric, datapoint in datapoints:
       if isinstance(datapoint[1], float):
         value = ("%.10f" % datapoint[1]).rstrip('0').rstrip('.')


lib/carbon/client.py
```diff
@@ -479,6 +479,7 @@ def clientProtocol(self):
 class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
 
   def _sendDatapointsNow(self, datapoints):
+    self.delimiter = b'\r\n'
```
ploxiln, line 482: This can just be like:

```python
class CarbonLineClientProtocol(CarbonClientProtocol, LineOnlyReceiver):
    delimiter = b'\r\n'
```

That's how carbon already does it for MetricLineReceiver:

```python
class MetricLineReceiver(MetricReceiver, LineOnlyReceiver):
  plugin_name = "line"
  delimiter = b'\n'
```

oddly, it looks like `b'\r\n'` is the default anyway?

https://github.com/twisted/twisted/blob/twisted-20.3.0/src/twisted/protocols/basic.py#L435
https://github.com/twisted/twisted/blob/twisted-19.10.0/src/twisted/protocols/basic.py#L435

pkruk: Thanks for your answer and yes you're right, default is b'\r\n' and this is quite strange. I will ask brindaveira in the issue about that, maybe there are some issues with using old carbon, or having an old library cached locally. So in that case this PR is unnecessary :)

pkruk:
I will close it for now, because it's not needed.

Issue #836, opened 2019-01-11
Improve carbon performance for non-tagged series


deniszh:
Related to https://github.com/graphite-project/carbon/issues/816
We can skip updates for tagdb when series with no tags - https://github.com/graphite-project/carbon/issues/816#issuecomment-431408473

deniszh:
I'll probably try to attack that.

deniszh:
For the record - it's not that easy. For now disabling TagDB is probably only option.

deniszh:
Fix implemented in #903

PR #903, opened 2020-10-04
Improve carbon performance for non-tagged series

deniszh:
Fixing #836 
Introducing new config variable `SKIP_TAGS_FOR_NONTAGGED`, which is True by default. When enabled it will avoid to add metrics to tag queue. TBH I see no usecases when it could be useful to disable such behaviour, but I prefer to not proceed with non-configureble behaviour changes.
Also switching tests to python 3.8, adding pypy3, and skipping broken test on MacOS


DanCech (approved):
Seems reasonable, lately I've actually been questioning the decision to put untagged metrics into the tag index since it causes the "name" tag to fill up with all the untagged series names.

PR #907, opened 2020-10-16
Fixes #906 (BAN-B101) Assert statement used outside of tests

### LINKED ISSUE #906: (BAN-B101) Assert statement used outside of tests
## Description
Usage of `assert` statement in application logic is discouraged. `assert` is removed with compiling to optimised byte code (`python -o` producing *.pyo files). Consider raising an exception instead. Ideally, `assert` statement should be used only in tests. Tip: Make sure `test_patterns` are defined in `.deepsource.toml` to avoid false-positives - https://deepsource.io/docs/config/deepsource-toml.html#test-patterns

## Occurrences
There are 4 occurrences of this issue in the repository.

See all occurrences on DeepSource &rarr; [deepsource.io/gh/graphite-project/carbon/issue/BAN-B101/occurrences/](https://deepsource.io/gh/graphite-project/carbon/issue/BAN-B101/occurrences/)

rohankhanna:


lib/carbon/storage.py
```diff
@@ -136,10 +136,12 @@ def loadAggregationSchemas():
     try:
       if xFilesFactor is not None:
         xFilesFactor = float(xFilesFactor)
-        assert 0 <= xFilesFactor <= 1
+        if not (0 <= xFilesFactor <= 1):
```
deniszh, line 139: Unnecessary parentheses after keyword PYL-C0325 https://deepsource.io/gh/graphite-project/carbon/run/45eefeec-464d-4e8b-987c-127ceaaa7f4a/python/PYL-C0325

lib/carbon/storage.py
```diff
@@ -136,10 +136,12 @@ def loadAggregationSchemas():
...
+          raise AssertionError("xFilesFactor value out of [0,1] bounds")
       if aggregationMethod is not None:
         if state.database is not None:
-          assert aggregationMethod in state.database.aggregationMethods
+          if not (aggregationMethod in state.database.aggregationMethods):
```
deniszh, line 143: Unnecessary parentheses after keyword PYL-C0325 https://deepsource.io/gh/graphite-project/carbon/run/45eefeec-464d-4e8b-987c-127ceaaa7f4a/python/PYL-C0325

deniszh:
Hi @rohankhanna 

Could you please fix PYL-C0325 introduced in your change?

Thanks!

rohankhanna:
@deniszh Is there anything else I should do? Is this good to merge?

lib/carbon/hashing.py
```diff
@@ -89,7 +89,8 @@ def remove_node(self, key):
     self.ring_len = len(self.ring)
 
   def get_node(self, key):
-    assert self.ring
+    if not self.ring:
+      raise AssertionError("self.ring not set")
```
piotr1212, line 93: I'd move this check to the end of the constructor. No point checking this on every get_node(). The whole idea of assertions is that they are optimised away in production code but you can use it while developing/debugging.

piotr1212 (changes requested):

lib/carbon/hashing.py
rohankhanna: I didn't realize that we were deliberately trying to optimise away assertions.

piotr1212: This code is 10 years old, no clue what was the intention. I also have no clue why Deepsource thinks having assertions in code outside of tests is a bad thing. I personally don't use them but there are many reasons one might want to, there is many CS research on that topic.

Anyway this looks good to me.

piotr1212: Ah, but now tests fail... Aaargh

piotr1212: We can probably just remove the whole check, it is kinda pointless, if ring = None then the bisect_left() fails anyway.

piotr1212 (approved):

Issue #905, opened 2020-10-16
(PYL-W0613) Function contains unused argument


rohankhanna:
## Description
An unused argument can lead to confusions. It should be removed. If this variable is necessary, name the variable `_` or start the name with `unused` or `_unused`.

## Occurrences
There are 5 occurrences of this issue in the repository.

See all occurrences on DeepSource &rarr; [deepsource.io/gh/graphite-project/carbon/issue/PYL-W0613/occurrences/](https://deepsource.io/gh/graphite-project/carbon/issue/PYL-W0613/occurrences/)

PR #911, opened 2020-10-24
Remove unused args #905

pkruk:
Like in this issue: https://github.com/graphite-project/carbon/issues/905

I runned tests, tox and looks fine :)


piotr1212:
These are callback functions, did you check if they are not called with argument from the reactor? And this code is not covered by unittests, unfortunately passing tests don't say much. You can check test coverage by running `coverage html`, it then generate an html report in the `htmlcov` directory.

deniszh:
Yes, what @piotr1212 said. I would not touch callback functions without very important reason.

pkruk:
you're right, then i will drop the changes which touch a callback functions :)

PR #945, opened 2023-07-08
Add Python 3.12 Support

damaestro:
- Remove use of imp
 - Update tests for unittest changes
 - Update ConfigParser read

### DIFF
--- a/lib/carbon/routers.py
+++ b/lib/carbon/routers.py
@@ -1,4 +1,3 @@
-import imp
 from carbon.hashing import ConsistentHashRing, carbonHash
 from carbon.util import PluginRegistrar
 from six import with_metaclass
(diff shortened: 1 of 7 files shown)


deniszh:
Hi @damaestro 

Could you please rebase against latest changes in master? I fixed the tests, should be green now.

lib/carbon/routers.py
damaestro, line 1: The porting to importlib is incomplete and needs to be reviewed by a maintainer that knows how this should behave.

imp | importlib
-- | --
imp.get_suffixes() | importlib.machinery.SOURCE_SUFFIXES, importlib.machinery.EXTENSION_SUFFIXES, and importlib.machinery.BYTECODE_SUFFIXES
imp.load_module() | importlib.import_module()

https://docs.python.org/dev/whatsnew/3.12.html#removed

The function signature has changed and the call needs to be cleaned up: https://docs.python.org/dev/library/importlib.html#importlib.import_module

damaestro: Use of `imp` was only for `carbon-client` and that was removed https://github.com/graphite-project/carbon/commit/1ebd4c16cf62bcf6630d77ba593f3f937d278951

I've removed the function related to this functionality.

deniszh:
Didn't trigger tests for some reason. I redid tests in #951, LGTM, will merge.

deniszh:
Thank you, @damaestro, merged with master.

Issue #922, opened 2021-08-17
Python 3.9.6 causes warning about buffering for binary mode


wvolz:
Running Carbon 1.1.8 is causing a warning when starting up under Python 3.9.6:

`lib/python3.9/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used`

I believe the issue is caused by the "1" on this line:
https://github.com/graphite-project/carbon/blob/9fad18df5731271aab6f5c81d32eddcecdc1a695/lib/carbon/log.py#L24

JohnHay:
I also see it on FreeBSD with python 3.8.12:

```
Starting carbon.
Starting carbon-cache (instance a)
/usr/local/lib/python3.8/os.py:1023: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
```

RedShiba:
catching the same issue

PR #953, opened 2024-02-07
fix log.py warning about buffering

earthgecko:
This is a fix for  Python 3.9.6 causes warning about buffering for binary mode #922, setting this to 0 is the default behaviour and results in the default buffer size being used.

The warning was released in Python 3.8.0 and is still present in 3.10.13 (tested, works and logging works fine) and is still present in 3.12.2 (https://github.com/python/cpython/blob/v3.12.2/Lib/_pyio.py#L231)

The version check may be excessive because who would be running < 3.8.0, but you never know.


deniszh (approved):

deniszh (approved):
Smh missed that, looks good

Issue #956, opened 2024-08-21
[BUG] Even with flow control enabled, carbon aggregator drops incoming metrics when output queues are full


bucko909:
There is a config setting, `USE_FLOW_CONTROL`, which defaults to enabled. The stated behaviour is that, when enabled, carbon will pause reading metrics when its sink is full. Indeed, carbon _does_ pause receiving metrics, but it also drops any "in-flight" metrics.

**Test setup**

Set up Carbon to not relay aggregator misses, and to use line protocol (so it's easy to test from bash). Also set up `a.*` to aggregate every 3 seconds, and carbon to spit out stats every 1 second, just to avoid waiting around. And verify that `FLOW_CONTROL` is `True`, so that data shouldn't be dropped when buffers fill.

```diff
~/carbon$ (cd test-root/conf; for i in *.conf; do diff -u0 ../../conf/$i.example $i; done)
--- ../../conf/aggregation-rules.conf.example   2021-10-08 18:21:23.195248379 +0100
+++ aggregation-rules.conf      2024-08-21 13:31:47.789993319 +0100
@@ -59,0 +60,2 @@
+
+a.<a> (3) = sum a.<a>
--- ../../conf/carbon.conf.example      2024-07-08 17:43:30.486176214 +0100
+++ carbon.conf 2024-08-21 14:35:32.558819082 +0100
@@ -436 +436 @@
-DESTINATIONS = 127.0.0.1:2004
+DESTINATIONS = 127.0.0.1:2003
@@ -441 +441 @@
-# DESTINATION_PROTOCOL = pickle
+DESTINATION_PROTOCOL = line
@@ -614 +614,2 @@
-DESTINATIONS = 127.0.0.1:2004
+DESTINATIONS = 127.0.0.1:2003
+DESTINATION_PROTOCOL = line
@@ -668 +669 @@
-# CARBON_METRIC_INTERVAL = 60
+CARBON_METRIC_INTERVAL = 1
@@ -680,0 +682 @@
+LOG_AGGREGATOR_MISSES = False
~/carbon$ grep ^USE_FLOW_CONTROL test-root/conf/carbon.conf 
USE_FLOW_CONTROL = True
USE_FLOW_CONTROL = True
USE_FLOW_CONTROL = True
```

Fire up a socket to look at what comes out:

```
~/carbon$ nc -l 2003 | grep -i drop
```

Fire up an aggregator:

```
[...]

PR #958, opened 2024-08-21
Allow the aggregator to consistently accept past data

bucko909:
The discovery of #956 was due to replaying a huge swathe of past data into our aggregator instance and finding the data full of holes. I'd assumed it was due to early-expiry of old `IntervalBuffer` objects, but after writing this fix I found this wasn't the case. However, the behaviour is still likely incorrect when data is played over an aggregation interval. Suppose the aggregator config is:

```
a (60) = sum a
```

And the input is:

```
a 0 1
...
a 1 19
<aggregation timeout triggers here at time (MAX_AGGREGATION_INTERVALS + 1) * 60 or higher>
a 1 20
...
a 1 59
```

Then the aggregation will aggregate seconds 0-29 and output this (value 20) to the cache, then expire the interval as it's too old, then in the next timeout, aggregate seconds 31-60, output this to the cache and expire the interval again. The final value is the aggregate of seconds 20-59, which is 40 instead of the expected 60. There is no way to control when aggregation occurs, so replayed data will be inconsistent, and in this case could be any value between 1 and 60, though if the replay is faster than real-time, the most likely will be 60.

After these changes, we allow the system to keep the newest interval prior to the current time-frame, provided it's been _submitted_ within the current time-frame. This means if old data is submitted in-order, as would be expected from a replay, the intervals will always be aggregated in a complete state at least once before expiry. This is somewhat robust to curiosities in the timestamps of the real-time submitting process (for example, if that process happens to round time down, or be pushing a prediction of future data or similar), as the number of kept intervals is based on submission time rather than interval time.

### DIFF
--- a/conf/carbon.conf.example
+++ b/conf/carbon.conf.example
@@ -637,9 +637,16 @@ USE_FLOW_CONTROL = True
 # You shouldn't need to tune this unless you really know what you're doing.
 MAX_DATAPOINTS_PER_MESSAGE = 500
 
-# This defines how many datapoints the aggregator remembers for
-# each metric. Aggregation only happens for datapoints that fall in
-# the past MAX_AGGREGATION_INTERVALS * intervalSize seconds.
+# This defines how many intervals behind the current one the aggregator
+# remembers for each metric. Intervals are expired when either:
+#   * They have seen no new datapoints in the last
+#   (MAX_AGGREGATION_INTERVALS * configured_frequency) seconds, or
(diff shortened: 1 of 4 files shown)


bucko909:
Push just reset to preferred email.

deniszh:
@bucko909 : linter says that current_interval is not in use, could you please check?
Ignore Deepsource, it should be removed

deniszh:
Another great addition, thanks, @bucko909 !

PR #959, opened 2024-08-21
make metrics from carbon-cache go via relay config

bucko909:
This is necessary for any hashed sharding setup: The cache will write metrics to its local cache, not to the _correct_ cache for the hashing scheme. This means only datapoints synced to disk are available. Worse, in a highly available setup, metrics differ depending on which node you happen to be fetching from.


bucko909:
This obviously isn't an ideal patch, as it requires a valid relay config to exist in the cache. Our solution is to set up the relay to submit to the HA router port, hence the `ConstantRouter` in the solution. This is obviously optional, but might be useful for other reasons.

Probably this should be a config option, something like `CACHE_METRICS_VIA_RELAY = False`.

bucko909:
Update just splits out the `ConstantRouter`.

deniszh:
@bucko909 : tests should be amended for changed behaviour, could you please amend it? Thanks!

bucko909:
I can do, but I think I need agreement on what the changed behaviour should be -- what's here cannot be shipped as it'll break existing configs.

"Proper" approaches are:

* Simply to require a default-off config setting to enable this behaviour, though it'll then remain as a trap for anyone who tries to set up a HA setup and doesn't thoroughly read the config file. Perhaps it can emit a warning if there are multiple instances configured but this setting is turned off?
* Have the cache read the relay section of the config file and copy the relay's submission behaviour. Obviously this is quite confusing, as the cache won't work without the relay configured, and it'll probably break single-pipeline setups (which don't need this anyway).

I'll submit a patch with the former behaviour and a warning for now; you can decide if that's appropriate. The tests will then work fine if I do this, though you may also want to require a test for the new behaviour. I'll see if I can get one to work.

bucko909:
I've added the setting and some warnings in the config file. I'm not sure it'd be possible to add a warning about configuration easily, because one daemon needs to warn depending on another's config, which isn't really in the design spec of the configuration's loader.

Some caveats:

* This is untested in its current configuration, though hopefully clearly won't break any existing installation. I'll be back at work next week and can probably test this more thoroughly on one of our live deployments then, though I've had an extended absence so it's unlikely to be my first priority.
* The test for `ConstantRouter` is a bit of a hack. I think realistically I should probably just remove it -- it's equivalent to any hashing approach with `REPLICATION_FACTOR = len(DESTINATIONS)` anyway. I think my adding it was simply misguided.

Let me know if you think this is reasonable.

deniszh:
@bucko909 : sorry, somehow missed that. Looks good, will merge it!

